PR #1, opened 2021-06-08
use createDerived from nanostores

Guria:
not sure I am doing right thing, but worth to try :)

### DIFF
--- a/core/router/index.test.ts
+++ b/core/router/index.test.ts
@@ -3,15 +3,15 @@ import {
   RouteParams,
   createStore,
   getValue,
-  Page
+  Page,
+  ReadableStore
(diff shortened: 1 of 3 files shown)


Guria:
@ai does it makes sense?

ai:
Sorry. Had no time to look yet.

ai:
Looks good! I wrote this block before I finished Nano Stores, so it was out of current Nano Store philosophy.

Just a small question: why do we need `as const` in `Map`?

Guria:
@ai 
image
It might not needed in this task, but it enables more sctrict inferring.

We will get ` Set<"start" | "signin">` instead of ` Set<string>`

PR #87, opened 2024-03-20
Fix RSS/Atom feed detection

kossnocorp:
This commit fixes certain RSS/Atom feeds that were not detected because `firstChild` was a non-element node (`xml-stylesheet` in [my case]). This is because `firstChild` also returns `#text` and `#comment` nodes. To fix that, I used `firstElementChild`, which always returns element nodes.

See [MDN]:

> To avoid the issue with node.firstChild returning #text or #comment nodes, Element.firstElementChild can be used to return only the first element node.

[my case]: https://www.channelnewsasia.com/api/v1/rss-outbound-feed?_format=xml&category=10416
[MDN]: https://developer.mozilla.org/en-US/docs/Web/API/Node/firstChild

## Motivation

I wanted the app to detect [an RSS feed](https://www.channelnewsasia.com/api/v1/rss-outbound-feed?_format=xml&category=10416).

## Screenshot or video

<div>
    <a href="https://www.loom.com/share/d67b470c455649b09e4b9599ebe063b4">
      <p>RSS/Atom feed fix - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/d67b470c455649b09e4b9599ebe063b4">
      
    </a>
  </div>

## Checklist

  - [ ] Look for alternatives, not just take the most popular one.
  - [ ] Check alternatives for [bundle size](https://bundlejs.com/), [`node_modules` size](https://packagephobia.com/), and [subdependencies number](https://npmgraph.js.org).
  - [ ] Check that the project is active by their repository/issues/PR.
  - [ ] Run `cd web && pnpm size`, check the JS bundle size changes, about limits if necessary.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:


kossnocorp:
Peace and love, but this checklist is exhausting. It made me think **a lot**!

PR #88, opened 2024-03-20
Where Ask Questions -> Where to Ask Questions

toplenboren:
Fixes # (No issue)

I changed `Where Ask Question` to `Where to Ask Questions` in `CONTRIBUTING.md` file

## Motivation

Hi! I'm sorry to nitpick, but `where to ask questions` is not really grammatically correct.. 

## Checklist

  - [x] Look for alternatives, not just take the most popular one.
  - [x] Check alternatives for [bundle size](https://bundlejs.com/), [`node_modules` size](https://packagephobia.com/), and [subdependencies number](https://npmgraph.js.org).
  - [x] Check that the project is active by their repository/issues/PR.
  - [x] Run `cd web && pnpm size`, check the JS bundle size changes, about limits if necessary.
- [x] Think about testing
- If you changed web client:
- If you changed `core/`:

## Thanks!

Thanks for the project and for the inspiring vision of local first web!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 # Contributing
 
 - [Language](#language)
-- [Where To Ask Questions](#where-to-ask-questions)
+- [Where to Ask Questions](#where-to-ask-questions)
 - [Install Environment](#install-environment)
 - [Run Web Client](#run-web-client)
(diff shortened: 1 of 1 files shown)


ai:
Thanks!

PR #90, opened 2024-03-20
Add StackBlitz sandbox link

sashachabin:
Let's add [StackBlitz button](https://developer.stackblitz.com/guides/integration/open-from-github) to README.md.

Contributors can quickly try out [slowreader in sandbox](https://stackblitz.com/fork/github/hplush/slowreader?file=web/pages/start.svelte) without `git clone` and instructions

Link contains:
- /fork — open repo fork, everything is saved without refresh
- ?file=web/pages/start.svelte — opens file with 1st interface

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,8 @@ You can ask any question to the maintainer in Telegram: **[`@sitnik`](https://t.
 
    - But for one-time work, you can manually install [Node.js](https://nodejs.org/en/download) and [pnpm](https://pnpm.io/installation) versions according to [`.tool-versions`](./.tool-versions) file.
 
+   - For an initial overview and run the project without downloading, you can try the online sandbox on [StackBlitz](https://stackblitz.com/fork/github/hplush/slowreader?file=web/main/main.svelte).
+
 2. Then install all npm dependencies by `pnpm`:
 
    ```sh
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Web app to combine feeds from social networks and RSS and to help read more meaningful and deep content.
 
-[![Open in StackBlitz](https://developer.stackblitz.com/img/open_in_stackblitz.svg)](https://stackblitz.com/fork/github/hplush/slowreader?file=web/main/main.svelte)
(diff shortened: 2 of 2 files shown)


ai:
1. Let’s also add link to `Install Environment` step as a first or second option. For all 3 options we need to explain when use  StackBlitz, when local, when manual.
2. Can we use HTML markup with inline size of the button? I know that they can change the size of the button, but page jumps during image loading is very painful for users.
3. What do you think if we will open `/web/main/main.svelte` since it is an entry point? With `Cmd+click` user will be able to go to any other page from that file.

toplenboren:
Awesome! Just wanted to add an Issue, that there are no docs on how to actually make `slowreader` up and running

Wanted to check out final product, but couldn't find a way :-(

ai:
> Wanted to check out final product, but couldn't find a way :-(

There is no final product 😅 this repo is just on tech prototype stage.

toplenboren:
> There is no final product 😅 this repo is just on tech prototype stage.

Even tech prototype deserves a place in web 😃

sashachabin:
> 1 Let’s also add link to Install Environment step as a first or second option. For all 3 options we need to explain when use StackBlitz, when local, when manual.

I tried adding the item to the installation section.

So it may be worth moving it  *after* these steps:

<details>
<summary>Alternative</summary>

image
</details>

<br />

> 2 Can we use HTML markup with inline size of the button? I know that they can change the size of the button, but page jumps during image loading is very painful for users.

Done 👍

<br />

> 3 What do you think if we will open /web/main/main.svelte since it is an entry point?

Good. We still close the first `start.svelte` almost immediately by pressing the button.

<br />

> With Cmd+click user will be able to go to any other page from that file.

Not working for Svelte in StackBlitz :(

ai:
> Not working for Svelte in StackBlitz :(

😭 Hope they will fix it somewhen later

> I tried adding the item to the installation section. It might be worth moving it after instructions as quote:

People will see it only when they finish the installing system locally and do not need it anymore.

Why not something like:

```md
1. Install envirovnment:
   - To fix small bug you can run web client directly in your browser with StackBlitz.
   - To start more complex task we recommend install `asdf`…
   - You can also manually install Node.js and pnpm …
```

ai:
Thanks. Amazing suggestion. Didn’t know about that button.

PR #112, opened 2024-03-26
Merge :root rules into the single css-rule

### LINKED ISSUE #99: Merge `:root` rules
Add custom PostCSS plugin to web client to merge all rules with `:root` selector into the single rule in the beginning of the file.

Note, that you should not take `:root` inside `@media` there.

Check `process:css` task in `web/` for an example of similar thing. Maybe we should integrate this code there (but rename files).

Konfuze:
Fixes #99

Adds custom PostCSS plugin to web client to merge all rules with :root selector into the single rule in the beginning of the file.

Plugin is used in the `web/scripts/clean-css.ts` script along with the `vars-cleaner` plugin because both plugins are processing build artifacts.

## Motivation

Consolidating all the `:root` rules should improve render time, reduce the bundle size, and make debugging easier

## Caveats

In theory, different order of rules in dev- and prod- environments could cause some hard-to-find bugs

## Checklist

- [x] Think about testing
- If you changed web client:
- If you changed `core/`:


web/test/vars-cleaner.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  return getVarsCleanerError()
+}
+
+beforeEach(() => {
+  resetCleanerGlobals()
+})
+
+test('clean unused palette colors', () => {
+  run(
```
ai, line 23: Let’s test return value here too.

web/test/vars-cleaner.test.ts
```diff
@@ -0,0 +1,62 @@
+import { equal } from 'node:assert'
```
ai, line 1: Thanks for tests, I like it 👍

web/scripts/clean-css.ts
```diff
@@ -1,29 +1,45 @@
-// Remove unused colors from palette (as CSS Custom Properties)
-// and throw and error if other CSS Custom Properties are unused.
+// Applies PostCSS plugins to clean each CSS file generated during the build.
```
ai, line 1: Now the main description is too general. User need to read 10 lines to find *why* we need this script.

What do you think if we move description here:

```js
// 1. Merge all rules with :root selector into the single rule
// in the beginning of the file
// 2. Remove unused colors from palette (as CSS Custom Properties)
// and throw and error if other CSS Custom Properties are unused.
```

And then we can remove `PLUGINS` constant and use `let cssCleaner = postcss([rootsMerger, varsCleaner])`

web/postcss/roots-merger.ts
```diff
@@ -0,0 +1,50 @@
...
+        if (rootRule.nodes.length > 0) {
+          root.prepend(rootRule)
+        }
+      },
+      Rule(cssRule) {
+        if (cssRule.selector !== ':root') {
+          return
+        }
+
+        if (cssRule.parent && cssRule.parent.type === 'atrule') {
```
ai, line 29:
```suggestion
        if (cssRule.parent?.type === 'atrule') {
```

A new JS syntax to improve readability

web/postcss/roots-merger.ts
```diff
@@ -0,0 +1,50 @@
...
+          return
+        }
+
+        cssRule.nodes.forEach(node => {
```
ai, line 33: You can use [`cssRule.walkDecls`](https://postcss.org/api/#rule-walkdecls)

web/postcss/roots-merger.ts
```diff
@@ -0,0 +1,50 @@
...
+        rulesToRemove.forEach(cssRule => cssRule.remove())
+
+        let rootRule = rule({ selector: ':root' })
+
+        rootRule.append(...rootNodes.values())
+        if (rootRule.nodes.length > 0) {
+          root.prepend(rootRule)
+        }
+      },
+      Rule(cssRule) {
```
ai, line 24: We are in the context of PostCSS plugin, let’s use just `rule` since everything here is about `css`

web/postcss/roots-merger.ts
```diff
@@ -0,0 +1,50 @@
...
+
+          // remove rule from the map to preserve the rules order
+          if (rootNodes.has(node.prop)) {
+            rootNodes.delete(node.prop)
+          }
+
+          rootNodes.set(node.prop, node)
+        })
+
+        rulesToRemove.push(cssRule)
```
ai, line 46: Why not remove it here?

web/postcss/roots-merger.ts
Konfuze: It's because we already have file-wide `rule` and I think it's a bad idea to rename it during the import:
`import { rule } from 'postcss'`

web/test/vars-cleaner.test.ts
Konfuze: I didn't notice that I had left an unused variable that should be returned with an error. I removed it and checked that we should get an undefined value if there are no errors

web/scripts/clean-css.ts
Konfuze: Great idea, I've updated the code

web/postcss/roots-merger.ts
Konfuze: ✅ done

web/postcss/roots-merger.ts
Konfuze: That's because we run both plugins at a time:
`const cssCleaner = postcss([rootsMerger, varsCleaner])`

If we remove the rule here, `varsCleaner` will not "see" it. This could cause the situation where unused rules will appear in the resulting CSS artifact even after processing by the `varsCleaner`.

We can run plugins one by one, but it seems that it would be more optimal to run them in the current way

web/postcss/roots-merger.ts
Konfuze: ✅ fixed

web/postcss/roots-merger.ts
ai: Using `import { rule } from 'postcss'` is prohibited in PostCSS 8, since your `import`’s PostCSS could be a different version (and AST will contain nodes from different PostCSS versions). It is not a big problem here, but a big problem for plugins published at npm.

The new way is `Rule(rule, { Rule })` — second argument has all classes to build AST.

web/postcss/roots-merger.ts
ai: Got it 🤔

web/postcss/roots-merger.ts
Konfuze: Aha, got it. ✅ fixed

ai:
Thanks. Amazing work.

Is it your account https://twitter.com/Konfuze ? I want to tell a little about that hack.

Konfuze:
> Is it your account https://twitter.com/Konfuze ? I want to tell a little about that hack.

Yes, that's me

PR #115, opened 2024-03-26
Improve atom loader w/ support for xml:base

### LINKED ISSUE #57: Add `xml:base` to Atom loader
Add [`xml:base`](https://www.w3.org/TR/xmlbase/) to link resolution in Atom loader.

Check the same attribute for RSS.

pekac:
Fixes #57 

Extended loader utils in `/core` to support `xml:base` tag for link resolution.

## Motivation

To add support for xml:base for link resolution.

## Checklist

- [x] Think about testing
- If you changed web client:
- If you changed `core/`:

### DIFF
--- a/core/loader/utils.ts
+++ b/core/loader/utils.ts
@@ -5,31 +5,26 @@ export function isString(attr: null | string): attr is string {
   return typeof attr === 'string' && attr.length > 0
 }
 
-function retrieveAllURLSegments(link: HTMLLinkElement): string[] {
+function buildFullURL(link: HTMLLinkElement, baseUrl: string): string {
   let href = link.getAttribute('href')!
   let urlSegments: string[] = [href]
   let parent: Element | null = link.parentElement
   while (parent) {
     let path = parent.getAttribute('xml:base') || ''
-    parent = parent.parentElement
     urlSegments.push(path)
+    parent = parent.parentElement
 
     if (path.startsWith('/') || path.startsWith('http')) {
       break
     }
   }
-  return urlSegments
-}
-
-function buildFullURL(link: HTMLLinkElement, baseUrl: string): string {
-  let urlSegments = retrieveAllURLSegments(link)
   return urlSegments.reduceRight(
-    (base, href) => new URL(href, base).href,
+    (base, url) => new URL(url, base).href,
     baseUrl
   )
 }
 
-export function findLink(text: TextResponse, type: string): string[] {
+export function findLinks(text: TextResponse, type: string): string[] {
   return [...text.parse().querySelectorAll('link')]
     .filter(
       link =>
@@ -44,8 +39,8 @@ export function hasAnyFeed(
   found: PreviewCandidate[]
 ): boolean {
   return (
-    findLink(text, 'application/atom+xml').length > 0 ||
-    findLink(text, 'application/rss+xml').length > 0 ||
+    findLinks(text, 'application/atom+xml').length > 0 ||
+    findLinks(text, 'application/rss+xml').length > 0 ||
     // TODO: Replace when we will have more loaders
     // found.some(i => i.loader === 'rss' || i.loader === 'atom')
     found.length > 0
--- a/core/loader/utils.ts
+++ b/core/loader/utils.ts
@@ -5,57 +5,28 @@ export function isString(attr: null | string): attr is string {
   return typeof attr === 'string' && attr.length > 0
 }
 
-type UrlSegmentType = 'relative' | 'absolute' | 'explicit'
-
-interface UrlSegment {
-  path: string
-  type: UrlSegmentType
-}
-
-function determineURLType(path: string): UrlSegmentType {
-  if (path.startsWith('/')) {
-    return 'absolute'
-  } else if (path.startsWith('https://') || path.startsWith('http://')) {
-    return 'explicit'
-  }
-  return 'relative'
-}
-
-function retrieveAllURLSegments(link: HTMLLinkElement): UrlSegment[] {
+function retrieveAllURLSegments(link: HTMLLinkElement): string[] {
+  let href = link.getAttribute('href')!
+  let urlSegments: string[] = [href]
   let parent: Element | null = link.parentElement
-  const urlSegments: UrlSegment[] = []
-
   while (parent) {
-    const path = parent.getAttribute('xml:base')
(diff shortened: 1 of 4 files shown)


pekac:
@ai I was thinking about 2 different approaches:
1. Create a new method, that would traverse the doc in one pass and retrieve both `links` and `xml:base`. 
2. Add another method, that for the found link(s) would determine the full url determined by `xml:base` tags.  

Approach 1) 
Seemed like an overkill and it went against the stylistic approach in code.
Performance gain is there but I'd argue it's not enough (at least for now) for the hustle it introduces with custom code for tree traversal and all the overhead needed to handle it.

Approach 2)
Went for this one. With it, we do get a nested loop (for each link we travel the parent chain) but at least complexity is not O(n^2) but closer to O(nlogn) since we are moving up the tree levels.

A request:
- Could you please assign the #57 to me? It's my first contribution to the open-source world :grin: 

A suggestion:
- Would suggest to rename the `findLink` function to `findLinks` since it already retrieves an "array" of links and not a "single" one. I believe it would communicate the purpose of the function more clearly, however didn't want to change it before you approve it.

Looking forward to the review!

core/loader/utils.ts
```diff
@@ -5,14 +5,67 @@ export function isString(attr: null | string): attr is string {
   return typeof attr === 'string' && attr.length > 0
 }
 
+type UrlSegmentType = 'relative' | 'absolute' | 'explicit'
```
ai, line 8: What do you think of using TypeScript `enum` instead?

The only downside, is `enum`’s is hard to debug in JS env, but it should not be a problem for internal code.

core/loader/utils.ts
```diff
@@ -5,14 +5,67 @@ export function isString(attr: null | string): attr is string {
...
+  return baseUrl
+}
+
+function buildFullURL(link: HTMLLinkElement, baseUrl: string): string {
+  const urlSegments = retrieveAllURLSegments(link)
+  baseUrl = buildBaseURLFromSegments(urlSegments, baseUrl)
+  return new URL(link.getAttribute('href')!, baseUrl).href
+}
+
 export function findLink(text: TextResponse, type: string): string[] {
```
ai, line 61: Sure, let’s rename to `findLinks`

core/loader/utils.ts
pekac: Agreed! 

Wanted to go with `enum` here but did a quick search in the entire repo and no results were found so I went with this.
Will update :+1:

core/loader/utils.ts
```diff
@@ -5,14 +5,67 @@ export function isString(attr: null | string): attr is string {
...
+    }
+  }
+  return urlSegments
+}
+
+function buildBaseURLFromSegments(
+  urlSegments: UrlSegment[],
+  baseUrl: string
+): string {
+  while (urlSegments.length > 0) {
```
ai, line 48: Wow, `xml:base` can be nested? 🥲

What do you think if will ignore nested `xml:base`? I think it will be extremely rare. I tried to look for `xml:base` use cases ([1](https://rssweblog.com/?guid=20050627094747), [2](https://www.odata.org/documentation/odata-version-2-0/atom-format/), [3](https://www.intertwingly.net/wiki/pie/PaceXmlBaseEverywhere), [4](https://github.com/Munter/netlify-plugin-checklinks/issues/566), [5](https://github.com/Ranchero-Software/NetNewsWire/issues/3662)) and all used absolute URL in the closes `xml:base`.

core/loader/utils.ts
ai: Feel free to fix another cases when we can use `enum`

core/loader/utils.ts
pekac: Yea, I cried a bit as well finding that out :sweat_smile: 

Essentially there are 3 types of path segments that can go in the `xml:base`:
- absolute ( /example/ )
- explicit ( http(s)://example.com* )
- relative ( example, example/, ./example, ../example, etc...)

Absolute segments - they end the nesting and are resolving to the document base url.
Explicit segments - they end the nesting and resolve to it as the root.
Relative segments - stack until they reach one of the first 2.

If we ignore the nesting, I think that in *most* cases we will be fine.
Honestly, don't see how (and why) anyone would create the nested-routing hell. 

On the other hand, this logic handles all the cases, both "most-used" ones and all the edge ones.
If we let it say like this - and most links resolve anyway to the nearest "absolute" `xml:base` tag, we won't have any shenanigans in the traversing and stacking the route segments.

The parent-chain traversing is necessary either way, since `xml:base` may be found on any element above the link. 

Not sure which one to go with tbh. Leaning just a bit towards covering everything since it's just a 3 functions and ~50lines of code for both rss and atom loaders and resolving to the nearest "absolute" `xml:base` will have the optimal control flow either way.

But anyway, it's your call.
If you think that's an overkill, I can just redo the logic to travel to the first "absolute" path segment and end it there.
In that case, would you like to handle "explicit" segments as well or just "absolute" ones?

core/loader/utils.ts
pekac: Do you have a preference from the code style perspective for `enum`s?

Does this look ok? Or you want them in all caps snake_case?

```typescript 
enum UrlSegmentType {
  Absolute = 'absolute',
  Explicit = 'explicit',
  Relative = 'relative'
}
```

ai: Looks good. Just let’s try to make identification a little bit shorter. `UrlSegmentType` → `UrlType`

core/loader/utils.ts
ai: I agree that it is cool to be ready for all cases. I just worry a little that the code is hard to read.

How much it will be simpler if we will cover only `absolute` case? If it will be significant simpler, let’s simplify. It will be still complicated, let’s keep it.

ai: (By the way, thanks for deep investigation 👍)

pekac: > I just worry a little that the code is hard to read.

Yea, my biggest concern was that as well. It's too hacky.

Let me sleep on it and see how far can I simplify for both approaches.

pekac: @ai , updated the PR. Should be much cleaner now.

ai: Yes, looks great

ai:
Thanks. Now we can parse much more Atom feeds.

Issue #98, opened 2024-03-22
Scroll page above on Next page click


ai:
On Fast posts page we have pagination and `Show next` button.

Now, after clicking on the button, page keeps scroll on the button of the page.

We need to fix it and scroll to top.

sanchopanda:
I want to take it

sanchopanda:
@ai I don't see the problem described. After clicking show next, scroll to the beginning of the feed https://monosnap.com/file/pE0u69zAycMun9GOhBTkbpu0kReSL1. Maybe I misunderstood the task?

ai:
Try to use another button “Mark as read …”

sanchopanda:
@ai At first I tried to implement it differently, i save it in [another branch](https://github.com/hplush/slowreader/compare/main...sanchopanda:slowreader:slowreader-98v2).  I  [assigned a link to an element to a key](https://github.com/hplush/slowreader/compare/main...sanchopanda:slowreader:slowreader-98v2#diff-53eedf36eb41eeda8138e6cad6ac43334af1bd2855ea068cb28bc228b833b432R12:~:text=setElement(%27stepOneElement%27%2C%20stepOne)) and [calling scrolling on the element by key directly in core/fast](elements.stepOneElement.scrollTo), but then I changed it to just dispatching event in window.

sanchopanda:
> @ai At first I tried to implement it differently, i saved it in [another branch](https://github.com/hplush/slowreader/compare/main...sanchopanda:slowreader:slowreader-98v2). I [assigned a link to an element to a key](https://github.com/hplush/slowreader/compare/main...sanchopanda:slowreader:slowreader-98v2#diff-53eedf36eb41eeda8138e6cad6ac43334af1bd2855ea068cb28bc228b833b432R12:~:text=setElement(%27stepOneElement%27%2C%20stepOne)) and [calling scrolling on the element by key directly in core/fast](https://github.com/hplush/slowreader/compare/main...sanchopanda:slowreader:slowreader-98v2#diff-53eedf36eb41eeda8138e6cad6ac43334af1bd2855ea068cb28bc228b833b432R12:~:text=elements.stepOneElement.scrollTo), but then I changed it to just dispatching event in window.

I already read the comments, forget it)

ai:
I just tried to read posts in https://dev.slowreader.app and scroll doesn’t work :(

sanchopanda:
Ok, i'll take a look a look

sanchopanda:
[...]

PR #125, opened 2024-03-27
fix(#98): Scroll page above on 'Mark as read and next load' click

sanchopanda:
Fixes https://github.com/hplush/slowreader/issues/98

Implemented scrolling to the top of the fast feed after clicking the mark read and load next button

### DIFF
--- a/core/fast.ts
+++ b/core/fast.ts
@@ -179,6 +179,9 @@ export async function markReadAndLoadNextFastPosts(): Promise<void> {
     await Promise.all($posts.get().map(({ post }) => deletePost(post.id)))
     if ($nextSince.get()) {
       await load(category, $nextSince.get())
+      if (window) {
+        window.dispatchEvent(new CustomEvent('loadNextFast'))
+      }
     } else {
       $posts.set([])
       $loading.set(false)
--- a/core/fast.ts
+++ b/core/fast.ts
@@ -179,7 +179,7 @@ export async function markReadAndLoadNextFastPosts(): Promise<void> {
     await Promise.all($posts.get().map(({ post }) => deletePost(post.id)))
     if ($nextSince.get()) {
       await load(category, $nextSince.get())
-      if (window) {
+      if (typeof window !== 'undefined') {
         window.dispatchEvent(new CustomEvent('loadNextFast'))
       }
     } else {
--- a/core/fast.ts
+++ b/core/fast.ts
@@ -179,9 +179,6 @@ export async function markReadAndLoadNextFastPosts(): Promise<void> {
     await Promise.all($posts.get().map(({ post }) => deletePost(post.id)))
     if ($nextSince.get()) {
       await load(category, $nextSince.get())
-      if (typeof window !== 'undefined') {
-        window.dispatchEvent(new CustomEvent('loadNextFast'))
-      }
     } else {
       $posts.set([])
       $loading.set(false)
--- a/core/refresh.ts
+++ b/core/refresh.ts
@@ -110,7 +110,7 @@ export async function refreshPosts(): Promise<void> {
             filters = prepareFilters(await loadFilters({ feedId: feed.id }))
           }
           for (let origin of posts) {
-            if (feedStore.deleted) {
+            if (feedStore.deleted || wasAlreadyAdded(feed, origin)) {
               await end()
               return
             }
--- a/core/refresh.ts
+++ b/core/refresh.ts
@@ -110,7 +110,7 @@ export async function refreshPosts(): Promise<void> {
             filters = prepareFilters(await loadFilters({ feedId: feed.id }))
           }
           for (let origin of posts) {
-            if (feedStore.deleted || wasAlreadyAdded(feed, origin)) {
+            if (feedStore.deleted) {
               await end()
               return
             }
--- a/web/pages/fast.svelte
+++ b/web/pages/fast.svelte
@@ -14,9 +14,17 @@
   import Loader from '../ui/loader.svelte'
   import PostCard from '../ui/post-card.svelte'
   import TwoStepsPage from '../ui/two-steps-page.svelte'
+
+  let twoStepsPageInstance: TwoStepsPage
+
+  fastPosts.subscribe(() => {
+    if ($fastPosts.length > 0) {
+      twoStepsPageInstance.scrollFirstToTop()
+    }
+  })
 </script>
 
-<TwoStepsPage title={$t.pageTitle}>
+<TwoStepsPage bind:this={twoStepsPageInstance} title={$t.pageTitle}>
   <div slot="one">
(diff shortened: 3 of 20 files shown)


core/fast.ts
```diff
@@ -179,6 +179,9 @@ export async function markReadAndLoadNextFastPosts(): Promise<void> {
     await Promise.all($posts.get().map(({ post }) => deletePost(post.id)))
     if ($nextSince.get()) {
       await load(category, $nextSince.get())
+      if (typeof window !== 'undefined') {
```
ai, line 182: We can't use `window` in the core since we will not have it on mobile.

web/ui/two-steps-page.svelte
```diff
@@ -5,16 +5,27 @@
...
+  function loadNextFastHandler() {
+    stepOneElement.scrollTo({
+      top: 0,
+      behavior: 'smooth'
+    })
+  }
+
   onMount(() => {
     document.title = title + ' › ' + prevTitle
+    window.addEventListener('loadNextFast', loadNextFastHandler)
```
ai, line 19: `<TwoStepsPage>` is universal component used in many pages.

Having a hack just for a specific page is not long-term solution because a few month later it will be full of hacks for different pages.

What do you think of we will move this code to `web/pages/fast.svelte`?

You can export method from two-side-page and call it in fast page on fast posts store changes.

sanchopanda: @ai Should we provide a method for scrolling in the second column? And if so, is it worth making a more universal method, like this, or just two methods, for the first column and for the second??

```
 let stepFirstElement: HTMLDivElement
 let stepSecondElement: HTMLDivElement

  export function scrollStepToTop(
    stepName: 'stepFirstElement' | 'stepSecondElement'
  ) {
    function scrollToTop(element: HTMLDivElement) {
      element.scrollTo({
        top: 0,
        behavior: 'smooth'
      })
    }

    switch (stepName) {
      case 'stepFirstElement':
        scrollToTop(stepFirstElement)
        break

      case 'stepSecondElement':
        scrollToTop(stepSecondElement)
        break

      default:
        scrollToTop(stepFirstElement)
        break
    }
  }
```

ai: Yes, it will be nice to be able to scroll both side.

But note, that `stepFirstElement` string will not be minified by builder.

It is better to export 2 different functions: `scrollFirstToTop()` and `scrollSecondToTop()`.

web/ui/two-steps-page.svelte
```diff
@@ -5,6 +5,26 @@
...
+    if (element instanceof HTMLElement) {
+      element.scrollTo({
+        top: 0,
+        behavior: 'smooth'
+      })
+    }
+  }
+
+  export function scrollFirstToTop() {
+    scrollToTop(stepFirstElement)
```
ai, line 21: Seems like if we will replace `scrollToTop` to `stepFirstElement?.scrollTo` we will make code simpler. The helper for the single function is not very useful and only makes system more complicated.

core/refresh.ts
```diff
@@ -110,7 +110,7 @@ export async function refreshPosts(): Promise<void> {
             filters = prepareFilters(await loadFilters({ feedId: feed.id }))
           }
           for (let origin of posts) {
-            if (feedStore.deleted || wasAlreadyAdded(feed, origin)) {
+            if (feedStore.deleted) {
```
ai, line 113: Why did we make this change?

web/pages/fast.svelte
```diff
@@ -14,9 +14,17 @@
   import Loader from '../ui/loader.svelte'
   import PostCard from '../ui/post-card.svelte'
   import TwoStepsPage from '../ui/two-steps-page.svelte'
+
+  let twoStepsPageInstance: TwoStepsPage
+
+  fastPosts.subscribe(() => {
```
ai, line 20: Every time you subscribe to some store think when you will unsubscribe from it.

Right now every time user visited `fast` page they will subscribe again and old subscription will be still active.

web/pages/fast.svelte
```diff
@@ -8,15 +8,29 @@
...
   import { getURL } from '../stores/router.js'
   import Button from '../ui/button.svelte'
   import Loader from '../ui/loader.svelte'
   import PostCard from '../ui/post-card.svelte'
   import TwoStepsPage from '../ui/two-steps-page.svelte'
+
+  let twoStepsPageInstance: TwoStepsPage
+
+  onMount(() => {
+    const unbindFast = fastPosts.subscribe(() => {
```
ai, line 22: You can make this code more compact by just returning `fastPosts.subscribe` result from callback.

core/refresh.ts
sanchopanda: Accidentally. I'm doing this for testing

web/pages/fast.svelte
```diff
@@ -8,15 +8,26 @@
...
   import { getURL } from '../stores/router.js'
   import Button from '../ui/button.svelte'
   import Loader from '../ui/loader.svelte'
   import PostCard from '../ui/post-card.svelte'
   import TwoStepsPage from '../ui/two-steps-page.svelte'
+
+  let twoStepsPageInstance: TwoStepsPage
+
+  onMount(() =>
+    fastPosts.subscribe(() => {
```
ai, line 22:
```suggestion
    return fastPosts.subscribe(() => {
```

Note, that in [PR check-list](https://github.com/hplush/slowreader/blob/main/docs/pull_request_template.md) we have an important rule:

>  Don’t rush. Check all changes in PR again.

web/pages/fast.svelte
```diff
@@ -8,15 +8,26 @@
...
   } from '@slowreader/core'
+  import { onMount } from 'svelte'
 
   import { getURL } from '../stores/router.js'
   import Button from '../ui/button.svelte'
   import Loader from '../ui/loader.svelte'
   import PostCard from '../ui/post-card.svelte'
   import TwoStepsPage from '../ui/two-steps-page.svelte'
+
+  let twoStepsPageInstance: TwoStepsPage
```
ai, line 19: Let’s make this variable shorter too. In other components I use just `element`.

web/pages/fast.svelte
sanchopanda: It seems to me that this is identical code, but is my version more compact? Just returning `fastPosts.subscribe` result from callback

ai: This is my bad, I missed that you do not have `{` in `onMount(() =>`.

Can we use:

```js
onMount(() => {
  return fastPosts.subscribe(() => {
```

If I read it wrong, somebody else can do it too.

ai:
Run `pnpm eslint --fix .` to fix all ESLint warnings.

Also, do not forget to run `pnpm test` before the review.

ai:
Done! Congrats with the first PR!

Issue #75, opened 2023-12-05
Add `theme-color`


ai:
We need a user with Chrome for Android to add this feature and test it.

- [ ] Add `theme-color` meta tag for light and dark color. We can use `--land-color` for that.
- [ ] Change them on switching to reading mode.
- [ ] Try to find a way to get colors in JS from CSS (maybe create separated CSS Custom Properties for background colors).

privorotskii:
I want to take it

privorotskii:
Hey @ai, I did some context gathering on that, can we discuss it a bit?

Do I get it right that the ultimate goal is to get the Safari's top bar color aligned to the theme color?

Here's a bit of context:

- As far as I see the meta tag is [handled by Safari, Chrome for Android and Samsung internet](https://caniuse.com/meta-theme-color)
  - Chrome for Desktop uses this tag only on PWA
- Safari on desktop uses this tag in compact mode which is set in settings, otherwise it's ignored
- Safari doesn't take into account the value of theme-color meta tag if it's set via `var()`
- Safari makes the top bar color aligned to the background by default itself
- In the reading mode it uses background color neither for page nor for the top bar

So to summarize: if the ultimate goal is to have the top bar color the same as the background, it's applied by Safari itself by default. 

However, we may still want to add this tag for the sake of making it explicit and to add it for Chrome Android and Samsung Internet, but at the moment I don't see a good way to do it without hardcoding the colours.
`--land-color` is a CSS custom property, so it should be something like this

```html
  <meta name="theme-color" content="var(--land-color)" />
```

But the `var` as a value is ignored by Safari, so it won't be applied

Safari Compact mode:
| Light mode | Dark mode |
|--------|--------|
| ![Image](https://github.com/hplush/slowreader/assets/10295457/b26692ac-db59-43b0-b356-8e129a8491aa)| ![Image](https://github.com/hplush/slowreader/assets/10295457/d3217c7b-2943-4e9d-b5b8-ba80ca35aed6) |

Safari separate mode:
| Light mode | Dark mode |
|--------|--------|
| ![Image](https://github.com/hplush/slowreader/assets/10295457/c410b3bd-9bb2-4163-a9fa-2fda6ebd5f4a)| ![Image](https://github.com/hplush/slowreader/assets/10295457/dd0f74d6-b711-4e49-8d83-26a08e602c77) |

Safari in reading mode:
| Light mode | Dark mode |
|--------|--------|
[...]

PR #126, opened 2024-03-27
Add <meta> tag for `theme-color` attribute

privorotskii:
Fixes [issue](https://github.com/hplush/slowreader/issues/75#issue-2025225827)
- Add <meta> `theme-color` tag
  - Update it's content based on `--land-color` CSS property

## Motivation
- The motivation is to add the styling of top bar for Chrome on Android
  - `note` Chrome on Android supports it only within light theme
  - `note` Safari also supports this tag, but it colors top bar without this tag by itself

## Screenshot or video

Chrome Android:

| Before | After |
|--------|--------|
| image | image |

Safari (no changes):

| Before | After |
|--------|--------|
| image | image | 
| image | image | 

## Checklist

- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:


web/main/colors.css
privorotskii, line 1: Most of the changes are made by linter, the intentional changes are adding `--theme-color`

privorotskii:
Hey @ai, I think I'm stuck here and need some help
So here's what I tried and why it didn't work out:

- `#idea` Just use the `var(--land-color)` as a <meta> tag content
  - `#issue` Neither Safari nor Chrome support var as a content. 
    - Which, probably, makes sense from some point of view — <meta> is HTML tag, HTML itself doesn't know about CSS functions
- `#idea` Use two <meta> tags, one for dark theme, one for light theme
  - `#note` We can do that using the media attribute, e.g. `<meta name="theme-color" media="(prefers-color-scheme: light)" content="..." />`
  - `#issue` We have not only dark or light themes, we also have different pages — slow and fast which have different --land-color property we'd like to use
- `#idea` Use JS to get the computed value of the --land-color
  - `#issue` The `--land-color` has value using `color-mix` function which isn't supported by this tag

So what I did in order to get the working version is taking the actual background colors via colorPicker and converting them to oklch.
I added those values as new CSS properties, but it doesn't feel like a good way to handle it:
- It reduces the sense of using color-mix for the main CSS variables since it's actually the same color, but in different notation which is harder to edit
- It's now needed to keep in mind to update this variable if the --land-color is updated, which is not obvious and could be missed at some point
- Not sure, but it could be probably different colors with --land-color, some discrepancies in the convertation process could arise

Does it make sense to do it this way or it's better to wait for better support from browsers? Or are there any better ideas? I'm open to any discussions and suggestions

web/main/helpers.ts
```diff
@@ -0,0 +1,13 @@
+export const actualizeThemeColor = () => {
```
ai, line 1: The helper is too small to move it to another file. Move it to `browser.ts`, it will be easier to analyze it.

web/main/helpers.ts
```diff
@@ -0,0 +1,13 @@
...
+  const landColor = getComputedStyle(document.documentElement).getPropertyValue(
```
ai, line 2: We are using `let` for variables (ESLint should print an error)

web/main/helpers.ts
```diff
@@ -0,0 +1,13 @@
...
+    '--theme-color'
+  )
+
+  const themeColorTag = document.querySelector('meta[name="theme-color"]')
```
ai, line 6: We can look for that tag only once to not spend resources every time.

web/main/helpers.ts
```diff
@@ -0,0 +1,13 @@
...
+
+  if (!themeColorTag || !landColor) {
```
ai, line 8: You will get shorter code by using

```js
if (themeColorTag && landColor) {
    themeColorTag.setAttribute('content', landColor)
}
```

web/main/colors.css
```diff
@@ -83,12 +89,14 @@
 .is-slow-theme {
   @media (prefers-color-scheme: light) {
     --land-color: color-mix(in oklab, var(--sand-100), var(--sand-200));
+    --theme-color: oklch(94.15% 0.01 82.4);
```
ai, line 92: Adding extra color variable is dangerous, because we may forget to update it.

What do you think if we will use:

```js
window.getComputedStyle(document.body).getPropertyValue('background') 
```

`getComputedStyle` returns styles after all CSS magic, so it will be just `rgb()` value, no custom properties, etc.

ai:
Good research. I suggested a way to avoid browsers limits.

web/main/colors.css
privorotskii: This works indeed (I changed it to get `background-color` though), good idea

web/main/browser.ts
```diff
@@ -30,6 +43,10 @@ if (!likelyWithKeyboard(window)) {
   root.classList.add('is-hotkey-disabled')
 }
 
+window.onload = () => {
+  actualizeThemeColor(themeColorTag)
+}
+
```
privorotskii, line 49: `question` Is it okay to add `window.onload` callback? Otherwise we get the `rgba(0,0,0,0)` at page loading, it will be updated to proper color only on e.g. changing tab

ai: Yes, just use `addEventListener('load', cb)`, since `window.onload` can be used by some other program or library.

web/main/browser.ts
```diff
@@ -11,15 +11,28 @@ import {
 
 import { locale } from '../stores/locale.js'
 
+const actualizeThemeColor = (themeColorTag: Element | null): void => {
+  let backgroundColor = window
+    .getComputedStyle(document.body)
+    .getPropertyValue('background-color')
+
+  if (themeColorTag && backgroundColor) {
+    themeColorTag.setAttribute('content', backgroundColor)
```
ai, line 20: Using var before it initialization is not good for readability.

Just move `actualizeThemeColor` below `let themeColorTag`

web/main/browser.ts
```diff
@@ -11,15 +11,28 @@ import {
...
+  }
+}
+
 let root = document.documentElement
+let themeColorTag = document.querySelector('meta[name="theme-color"]')
```
ai, line 25: Let’s use just `theme`. I prefer short variable names.

web/main/browser.ts
```diff
@@ -11,15 +11,28 @@ import {
 
 import { locale } from '../stores/locale.js'
 
+const actualizeThemeColor = (themeColorTag: Element | null): void => {
```
ai, line 14: Let’s use `updateTheme` to keep code style consistency of short names.

web/main/colors.css
ai: Strange. We use Stylelint to format CSS. If you have different formatting, you may use different linter.

web/main/browser.ts
privorotskii: Makes sense, I made it `themeTag` since we already have `theme` as an imported variable

web/main/browser.ts
privorotskii: Moved it below the let declaration, renamed the function param to `themeNode` — eslint complained on having the same name as the already declared variable above

privorotskii: Actually, there's one more option — we can make the function not accepting params and just working with the `themeTag` from the scope, does it sound good?

ai:
Next time,  don’t forget to run `pnpm test` before creating PR.

privorotskii:
@ai I did, it passes for me on this branch. Was something wrong after merging?

image

ai:
Size Limit was wrong.

https://github.com/hplush/slowreader/actions/runs/8507486659

Maybe after rebasing on new `main` branch.

PR #127, opened 2024-03-28
Add loader tests project

### LINKED ISSUE #96: Add loader tests on OPML file
We need a way to test that loaders are working on real websites.

The tool will test loaders by:
1. Developer open their RSS reader and save all subscriptions to OPML file
2. Then the developer run our script and pass the path to that OPML file
3. This script takes every RSS feed from OPML file and test that it could parse it and find it from website home

To finish the task:

- [ ] Add `loader-tests/` project to monorepo.
- [ ] Create `README.md` and example a little.
- [ ] Add project description to the main `README.md`
- [ ] Add script. Import `@slowreader/core` and call `setEnvrionment()` with Node.js APIs.
- [ ] Add npm script to call that script to main `package.json`. Document it in main `README.md`.
- [ ] Read arguments. Show help on no file in argument. Show nice errors if we can’t read file.
- [ ] Parse OPML and get list of RSS URLs.
- [ ] For each RSS try to find the right loader by calling `loader.isMineText()`. Show nice error if the loader was not found.
- [ ] For found loader call `loader.getMineLinksFromText()`. Be sure that it returned some posts.
- [ ] For URL remove path (`https://example.com/news.rss` → `https://example.com/`) and call use methods from `core/preview.ts` to be sure that reader will find that RSS URL from the home. If URL was not found show warning (not error, since sometimes RSS link is hidden).
- [ ] Add to `CONTRIBUTION.md` to `How to Find Next Task` step to use this tool if you have RSS reader.

Use colors output by `picocolors`. You can use `nanospinner` for loading spinner.

pekac:
Fixes #96 

Created a `loader-tests` project to create a testing tool for real-world opml files.

## Motivation

To show that our loaders actually work on real-world data and allow devs to use the tool on their files as well.

## Checklist

- [x] Think about testing
- If you changed web client:
- If you changed `core/`:


README.md
```diff
@@ -105,6 +105,7 @@ Slow Reader is a local-first app. Clients do most of the work, and the server ju
 - [`api/`](./api/): types and constants shared between clients and server.
 - [`docs/`](./docs/): guides for developers.
 - [`scripts/`](./scripts/): scripts to test project and configure Google Cloud. Check the script’s descriptions for further details.
+- [`loader-tests/`](./loader-tests/): loader testing tool for OPML files.
```
ai, line 108:
```suggestion
- [`loader-tests/`](./loader-tests/): integration tests for each social network or news format.
```

`loader` is our internal term. A new developer will not understand what it is.

Also it is not only for OPML files, since later we will do much more integration tests in that dir.

loader-tests/utils.ts
```diff
@@ -0,0 +1,50 @@
+import pico from 'picocolors'
```
ai, line 1: We can already use `util.styleText` from [the latest Node.js](https://nodejs.org/en/blog/release/v20.12.0#text-styling)

loader-tests/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@slowreader/loader-tests",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "start": "tsx ./index.ts"
+  },
+  "dependencies": {
```
ai, line 8: Let’s move all dependencies to `devDependencies`.

During the production build we will run `pnpm install --prod` to reduce security risks from dependencies. So we will need as minimum as possible dependencies for production deploy.

loader-tests/README.md
```diff
@@ -0,0 +1,3 @@
+# Slow Reader Loader tests
+
+Loader testing tool for OPML files.
```
ai, line 3: We need a little more details here. At least longer description as we have in `README.md` and explanation why we need this dir (to test that we can load posts from real Internet, not abstract unit tests).

Also, we will need a guide how to run OMPL test.

pekac: Sure - it's a draft PR, I actually have some questions but you are too fast :grin:

loader-tests/utils.ts
pekac: Not sure tho, I'm getting this error if I try to use it w/ `node v20.12.0`.
```
Module '"node:util"' has no exported member 'styleText'.ts(2305)
```

pekac:
So a couple of questions about the tasks #96.

First question:
> Add script. Import `@slowreader/core` and call `setEnvrionment()` with Node.js APIs.

I'm a bit confused here because the setupEnvironment function makes sense when we want to bind some logic in the store and use it from there for different platforms.
But `loader-tests` project won't have any business logic and will be used only for testing purposes.
Am I missing something?

Second:
I've defined a couple of fns in `utils` and in `index.ts` that we already have (or have something similar) in `core/loader` but are not exported. I wanted to double check if it was ok to export them since I didn't want to expose anything that's unnecessary. 
Fns like:  `isString` `getLoader` in `utils` and I'm using similar code to what we have in core for parsing xml files.
Leave separate but similar (or same) implementations or export stuff from core? (utils are hidden in core and some fns in preview.ts from core as well)

Third:
> For URL remove path (https://example.com/news.rss → https://example.com/) and call use methods from core/preview.ts to be sure that reader will find that RSS URL from the home. If URL was not found show warning (not error, since sometimes RSS link is hidden).

Here, are you referring to URLs to feeds defined in the OPML file or are you referring to specific posts URLs in the feed itself? 

Fourth:
Yes, I will improve the docs :grin:

loader-tests/utils.ts
pekac: @ai , it's probably `@types/node`, last release was 1w ago, and it's on `20.11.30`. It needs to catch up.

core/index.ts
```diff
@@ -21,3 +21,4 @@ export * from './post.js'
 export * from './i18n.js'
 export * from './html.js'
 export * from './slow.js'
+export { getTestEnvironment } from './test/utils.js'
```
ai, line 24: Let’s move it to `./environment.ts` then

pekac: Similar question for `./dom-parser.ts`

Is it ok to move it to `core` and to export it from there if it's needed?

We need it for loader-tests project and I believe in many more places soon.

ai: No, it is DOM polyfills hack. Let’s do it again in every project (I’m afraid to export polyfills since it could break environments by accidentally change globals).

pekac: Deal :+1:

CONTRIBUTING.md
```diff
@@ -116,12 +116,13 @@ Read our **[checklist for pull requests](./docs/pull_request_template.md)**. It
 If you are not just fixing a bug but want to join the core team, here is the path:
 
 1. Your first task is to test `pnpm test` in your environment and report issue.
-2. Go to our [Tasks plan](https://github.com/orgs/hplush/projects/1) and take next task from `Onboarding` column. If `Onboarding` is empty, you can look in `Ready to Take`.
+2. If you have a RSS Reader, your second task is to export all your subscriptions from your reader in a file. Then, check if our [loader-tests](./loader-tests/) project supports testing of your specific file type. If it does, run the `loader-test` script for your file and make sure all posts are returned correctly.
```
ai, line 119:
```suggestion
2. If you have a RSS Reader, your second task is to export all your subscriptions from your reader in a OPML file. Then run the `loader-test` script for your file and make sure all posts are returned correctly.
```

Let’s start from telling about OMPL-only export. If too many RSS readers are having different export name, we will think.

But it is very cool, that you think about `.xml` export.

core/environment.ts
```diff
@@ -118,3 +119,24 @@ export function getEnvironment(): Environment {
   }
   return currentEnvironment
 }
+
+const testRouter = atom<BaseRoute | undefined>()
+
+export function setBaseRoute(route: BaseRoute | undefined): void {
```
ai, line 125: `setBaseRoute` → `setBaseTestRoute`

Otherwise people may think that it is core router function.

loader-tests/README.md
```diff
@@ -0,0 +1,20 @@
...
+- `.xml`
+
+## Running the tool for OPML files
+
+- Check out [example.opml](./example.opml) for the structure of `.opml` files and testing purposes
+- Make sure to start the script from the monorepo root
+- Once in the root you can run:
+
+```sh
+$ pnpm feed-loader <path_to_your_file.opml>
```
ai, line 19:
```suggestion
pnpm feed-loader PATH_TO_YOUR_FILE.opml
```

In other examples we do not have `$` (I am open to add it, but to everywhere in separated PR).

loader-tests/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": "./",
+    "paths": {
+      "@slowreader/core": ["node_modules/@slowreader/core/index"]
```
ai, line 6: Should we add it to the global `tsconfig`?

loader-tests/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+}
+
+async function findRSSfromHome(feed: Feed): Promise<void> {
+  try {
+    keepMount(previewCandidates)
+    await setPreviewUrl(feed.htmlUrl)
+    await setTimeout(timerDurationInSec * 1000)
+    if (previewCandidates.get().length === 0) {
```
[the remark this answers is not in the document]
ai: Timeout code can be a little more complicated
https://advancedweb.hu/how-to-add-timeout-to-a-promise-in-javascript/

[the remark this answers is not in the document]
pekac: @ai sure, but check this out... if we just leave the fn like this:
```javascript 
function waitForTrue(store) {
  return new Promise(resolve => {
    let unbind = store.listen(value => {
      if (value === true) {
        unbind()
        resolve()
      }
    })
  }
}
```

and then use it like you said : ```javascript
waitForTrue(previewCandidatesLoading),
```

It will resolve immediately when *isLoading* is set from `false` to `true`, to indicate the loading process, not when the promise actually resolved.

So, the logic to wait for resolvement would have to be something like this:
```typescript
export function waitForStoreResolve(
  store: ReadableAtom<boolean>
): Promise<boolean> {
  return new Promise(resolve => {
    let loadingComparison = true
    let unbind = store.listen(value => {
      if (value && loadingComparison) {
        /**
         * value = true actually is an indicator of "loading in progress"
         * value changed from 'idle' to 'isLoading'
         * since
         *  */
        loadingComparison = false
      }

      if (!value && !loadingComparison) {
        /**
         * here it actually resolved, changed from 'isLoading' to 'done'
         * value = true -> value = false
         * it was loading and now is not anymore
         * */
        unbind()
        resolve(value)
      }
    })
  })
}
```

if this is not too hacky - I can include it.
I felt like it was too much so I just repeated what we did in the tests with the "stupid" timer.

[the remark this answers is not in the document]
ai: Yes, let’s go this direction. I will try to clean up code later.

loader-tests/tsconfig.json
pekac: Not sure. It's a project specific dep and not all projects use it :man_shrugging: 

If we put it in root `tsconfig` we also have to add it to root `package.json` deps so tsx can resolve it.
And if it's in root then `core` can also reference it? :dizzy:   

I'm leaning towards project-specific but it's your call. You want it in the root `tsconfig`?

ai: OK, let's keep it here. I maybe will do a research later

Issue #109, opened 2024-03-25
Avoid double `pnpm build` in CI tests


ai:
[`.github/workflows/test.yml`](https://github.com/hplush/slowreader/blob/main/.github/workflows/test.yml) contains 2 steps where we build web client assets: both in `test` and `staging`.

Let’s do it only in `test` and save as artifact to use in `deploy`. It will:
1. Speed up CI.
2. Increase security, since potentially malicious dependencies will not be located in a disk where we store Google Cloud credentials.

What to do:
1. Add step to upload `web/dist/` folder to artifact in `test` step. But do it only for `main` branch where we will run `staging`.
2. Add step to download artifacts in `staging` step.
3. Think about artifact cleaning. Seems like we will need them no more that for a single day.
4. Think about having multiple commits on CI in the same time. We need to be sure that only `staging` from commit A use artifacts from `test` job of commit A, not another commit.
5. Remove unnecessary steps from `staging` job. For instance, we can remove `Install dependencies` and `Build assets`. But let’s see what else.

pshergie:
I want to take the task

ai:
It was fast :D. It is on you.

pshergie:
> [.github/workflows/test.yml](https://github.com/hplush/slowreader/blob/main/.github/workflows/test.yml) contains 2 steps where we build web client assets: both in test and staging.
> 
> Let’s do it only in test and save as artifact to use in deploy

@ai I can see the `pnpm build` command in [staging](https://github.com/hplush/slowreader/blob/main/.github/workflows/test.yml#L50-L51) but not in `test`, am I missing something?

ai:
@pshergie in `test` we call [`pnpm test`](https://github.com/hplush/slowreader/blob/main/package.json#L11) which run all `test:` prefixed task, which run [`test:build`](https://github.com/hplush/slowreader/blob/main/web/package.json#L15) in web, where we call `pnpm build`.

ai:
Fixed by
https://github.com/hplush/slowreader/pull/128
https://github.com/hplush/slowreader/commit/351b6051f66a092a15b32e7ad9ed96b819ac9963

We got 16% faster CI (20 seconds).

pshergie:
@ai thanks for the help, looking forward for the next opportunity to contribute :)

PR #128, opened 2024-03-28
add build artifacts step to ci/cd config

pshergie:
Fixes # (issue)

1. Added a step to upload web/dist/ folder to artifacts in the `test` step for the `main` branch
2. Added a step to download artifacts in the `staging` step.
3.  Removed unnecessary steps from the `staging` job

## Motivation

the PR is based on the issue https://github.com/hplush/slowreader/issues/109

## Checklist

- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,14 +4,14 @@ on:
     branches:
       - main
     paths-ignore:
-      - "./**/*.md"
-      - "./.vscode/**"
-      - "./.husky/**"
+      - './**/*.md'
+      - './.vscode/**'
+      - './.husky/**'
   pull_request:
     paths-ignore:
-      - "./**/*.md"
-      - "./.vscode/**"
-      - "./.husky/**"
+      - './**/*.md'
+      - './.vscode/**'
+      - './.husky/**'
 jobs:
   test:
     permissions:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,14 +4,14 @@ on:
     branches:
       - main
     paths-ignore:
-      - './**/*.md'
-      - './.vscode/**'
-      - './.husky/**'
+      - "./**/*.md"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -29,6 +29,13 @@ jobs:
         run: pnpm test
         env:
           FORCE_COLOR: 1
+      - name: Upload artifacts
```
ai, line 32:
```suggestion
      - name: Upload web assets
```

Artifact is abstract term. We can be more precise.

.github/workflows/test.yml
```diff
@@ -43,12 +50,10 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-      - name: Install tools from asdf config
-        uses: ai/asdf-cache-action@v1
-      - name: Install dependencies
-        run: pnpm install --ignore-scripts
-      - name: Build assets
-        run: cd web && pnpm build
+      - name: Download built assets artifacts
```
ai, line 53:
```suggestion
      - name: Download web assets from test step
```

Let’s explain how made it for better readability

ai:
Looks good. Do you plan to add anything before removing Draft marker from PR?

pshergie:
Didn't expect you to respond that quickly :) I'm not 100% sure that this change will work correctly but I assume we won't know until the branch is merged to the master. I do not have any additional change in mind. It's draft because I am going to have limited availability this weekend

ai:
Looks great. Let's try to merge it.

ai:
Nope 😥

https://github.com/hplush/slowreader/actions/runs/8474691810/job/23221580518

ai:
The fix: https://github.com/hplush/slowreader/commit/351b6051f66a092a15b32e7ad9ed96b819ac9963

PR #129, opened 2024-03-29
Add Markdown linter

### LINKED ISSUE #97: Add Markdown linter
We need a linter to test Markdown files for:

1. That TOC has right titles and URLs.
2. That all links is alive.
3. That titles have the correct word's capitalization (`a`/`the` is small, noun is starting from big latter).
4. That file use curvy typography `"`→`”`, `'`→`’`.

To finish the task:
- [ ] Research for good Markdown linter. It should be able to check everything we need, be flexible and [use minimum size in node_modules](https://packagephobia.com/).
- [ ] Add `test:markdown` script to main `package.json`. So this script will be called on `pnpm test`.
- [ ] Find way to avoid calling in `.github/workflows/root.yml`.
- [ ] Add `.github/workflows/docs.yml` to run it on `**/*.md` files changes. 
- [ ] Check how fast it is. If it is very fast, add it to `nano-staged.json`.

ilyatitovich:
Fixes #97

When searching for a Markdown linter, I settled on [remark](https://github.com/remarkjs/remark). The linter has an extensive collection of plugins and presets for processing Markdown files, allowing us to check only the necessary rules. It is possible to write custom rules to perform specific tasks.

I added 4 packages:

- [remark-cli](https://www.npmjs.com/package/remark-cli): CLI linter interface
- [remark-lint-no-dead-urls](https://www.npmjs.com/package/remark-lint-no-dead-urls): plugin that checks whether links are live
- [remark-validate-links](https://www.npmjs.com/package/remark-validate-links): plugin that checks links to files and folders
- [remark-lint-code-block-style](https://www.npmjs.com/package/remark-lint-code-block-style): plugin that checks code blocks and can format the file if necessary. To format files during linting, you need to add `--output` to the script: `"test:markdown": "remark . --output"`.

We may also need the [remark-toc](https://github.com/remarkjs/remark-toc) plugin, which allows us to generate TOC for *.md files.

Additionally, a `.remarkrc.js` file has been added, in which the necessary plugins are connected.

Using this linter, we can check whether links in a document are alive. For other tasks we need to write our own rules.

If this linter suits us, then I’ll start to work with GitHub Actions to add .github/workflows/docs.yml to run linter on **/*.md files changes.

And I'm a little confused by the task: `Find way to avoid calling in .github/workflows/root.yml`, because I don't see this file in the `workflows` directory.

## Motivation

Add to the project a good, flexible linter for checking *.md files for mistakes.

## Checklist

- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:


.remarkrc.js
```diff
@@ -0,0 +1,12 @@
...
+
+const preset = {
+  plugins: [
+    'remark-lint-no-dead-urls',
+    'remark-validate-links',
+    'remark-lint-code-block-style'
+  ]
+}
+
+export default preset
```
ai, line 12: We can directly export object without saving it to variable

.remarkrc.js
```diff
@@ -0,0 +1,12 @@
+// File for setting up the markdown linter configuration.
```
ai, line 1: The descriptions should be ready for junior developers who don’t remember `linter` and `Markdown` words.

Something like:

```
The config of tool to check documentation in .md files for popular mistakes
```

package.json
```diff
@@ -19,6 +19,7 @@
     "test:focused": "tsx ./scripts/check-focused-tests.ts",
     "test:messages": "tsx ./scripts/check-messages.ts",
     "test:versions": "tsx ./scripts/check-versions.ts",
+    "test:markdown": "remark .",
```
ai, line 22: CI shows wrong warning

```
158:5-158:83 warning Cannot find file `(https:/github.blog/2019-02-14-introducing-draft-pull-requests/)` missing-file remark-validate-links:missing-file
```

Maybe it is a reason why PR is draft. Just note your.

.remarkrc.js
```diff
@@ -0,0 +1,12 @@
...
+// Add the required plugin to the plugins array.
+
+const preset = {
+  plugins: [
+    'remark-lint-no-dead-urls',
+    'remark-validate-links',
+    'remark-lint-code-block-style'
```
ai, line 8: Does it just check between <code>```</code> and indent code block format?

We do not need it since most users don’t know about indent based code blocks.

I told you about a different thing. If you have a code block in *the list* you need to add indent to <code>```</code> block.

Wrong:

<pre>
1. List item
```
code
```
3. List item
</pre>

Right:

<pre>
1. List item
   ```
   code
   ```
3. List item
</pre>

ai:
> We may also need the [remark-toc](https://github.com/remarkjs/remark-toc) plugin, which allows us to generate TOC for *.md files.

We may need to write custom rule here too, since we need not generate TOC but to test current TOC (i am not sure how this plugin works to be honest, maybe it checks).

> And I'm a little confused by the task: Find way to avoid calling in .github/workflows/root.yml, because I don't see this file in the workflows directory.

Oops. I renamed this file to `test.yml` since I created an issue.

Here is a problem. We run `test.yml` on [non-md changes](https://github.com/hplush/slowreader/blob/main/.github/workflows/test.yml#L6-L9) because it runs a lot of expensive test (like staging deploy).

We need to keep separation between “just test docs quickly on spellchecking PR” and “change a code”. But `test.yml` now runs *all tests* which started from `test:`. So we need to change `test.yml` to not run `test:markdown` (but run it on `pnpm test` which should test *everything*).

package.json
```diff
@@ -40,6 +41,10 @@
     "postcss-html": "1.6.0",
     "prettier": "3.2.5",
     "prettier-plugin-svelte": "3.2.2",
+    "remark-cli": "^12.0.0",
```
ai, line 44: Don’t forget to change Tools section in `README.md`

package.json
ai: Also try to make a mistake to docs and run `pnpm test:markdown`. Script should fail with error (not just print warning).

Try to break HTTP link (remove any symbol to make it dead), break local files link, etc.

ilyatitovich:
In order to exclude the running of the `"test:markdown"` script from GitHub Actions, I suggest adding the `grep -v "test:markdown"` filter. It will skip and run all scripts starting with `"test:"` except `"test:markdown"`.

Added `docs.yml` file to run tests specifically for .md files.

I removed the `remark-lint-code-block-style` package, because it really doesn’t do what we need. I have not yet discovered a plugin that checks indentation inside a list element.

Added `--frail` to the script: `"test:markdown": "remark . --frail"`. Now linter displays a warning and returns an error:

links


Added the remark tool to the **Tools** section in README.md.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,18 @@
+name: Markdown Tests
```
ai, line 1:
```suggestion
name: Markdown Linter
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,18 @@
...
+on:
+  push:
+    branches:
+      - main
```
ai, line 5: This workflow will not run on PR (and we need this linter to reduce review for PRs).

Copy this block from `test.yml` https://github.com/hplush/slowreader/blob/main/.github/workflows/test.yml#L10

.github/workflows/docs.yml
```diff
@@ -0,0 +1,18 @@
...
+    paths:
+      - '**/*.md'
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
```
ai, line 14: Very old action version

.github/workflows/docs.yml
```diff
@@ -0,0 +1,18 @@
...
+      - name: Install Markdown linter and plugins
+        run: pnpm install remark-cli remark-lint-no-dead-urls remark-validate-links --ignore-scripts
```
ai, line 16: Every time we will have different versions of `remark` here.

I suggest using pnpm lock file and just use the standard step with installing Node.js and all pnpm dependencies.

.github/workflows/test.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Install dependencies
         run: pnpm install --ignore-scripts
       - name: Run tests
-        run: pnpm test
+        run: pnpm test | grep -v "test:markdown"
```
ai, line 29: `grep -v` will just hide `test:markdown` output. But we will still spend CPU on Markdown linter and CI will still fails on Markdown issues (but without a proper output why).

.remarkrc.js
```diff
@@ -0,0 +1,6 @@
+// Configuration file used to set up a tool to find mistakes in .md files.
```
ai, line 1:
```suggestion
// Configuration file used to find popular mistakes in .md files.
```

Let’s make it shorter.

.remarkrc.js
```diff
@@ -0,0 +1,6 @@
...
+// Add the name of the plugin you want to use to the list of plugins.
```
ai, line 2: Other configs doesn’t have these description

```suggestion
```

ilyatitovich:
Updated the `docs.yml` file, adding PR bypass and standard installation of the environment and packages.

Updated the script in the `test.yml` file for running tests by adding the regular expression `'/^test:(?!markdown\b)/'`. It will allow us to run all tests except the "test:markdown". Checked locally - it works.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,29 @@
+name: Markdown Linter
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '**/*.md'
+  pull_request:
+    paths-ignore:
+      - './**/*.md'
```
ai, line 10: This workflow will ignore all PRs with `*.md` changes. Seems like it is a mistake.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+      - name: Install tools from asdf config
+        uses: ai/asdf-cache-action@v1
+      - name: Install dependencies
+        run: pnpm install --ignore-scripts
+      - name: Run tests
+        run: pnpm test:markdown
+        env:
+          FORCE_COLOR: 1
```
ai, line 29: We do not need `FORCE_COLOR` here. In `pnpm test` we need it because `pnpm` has an issue with nested tasks.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,29 @@
...
+    paths:
+      - '**/*.md'
+  pull_request:
+    paths-ignore:
+      - './**/*.md'
+      - './.vscode/**'
+      - './.husky/**'
+jobs:
+  test:
+    permissions:
```
ai, line 15: If we have a single `job`, let’s move `permission` to the root level as we have in another workflows with one job.

README.md
```diff
@@ -119,6 +119,7 @@ Global development tools:
 - [Prettier](./.prettierrc) to use the same code style formatting.
 - [TypeScript](./tsconfig.json) for strict type checking.
 - [ESLint](./eslint.config.js) to check for popular mistakes in JavaScript.
+- [remark](./.remarkrc.js) to find mistakes in .md files.
```
ai, line 122:
```suggestion
- [remark](./.remarkrc.js) to find mistakes in `.md` files.
```

1. Better formatting is better for readability.
2. File names should be in `<code>` according to practice in the web.
3. Without `<code>` the filename dot conflicts with typography dot.

ai:
1. When you will fix issues, press `Resolve conversation`
2. We need also to resolve conflicts

But I think it is the last code review. We are very close to merging.

ilyatitovich:
Corrected logical mistakes in `docs.yml`. Resolved conflicts.

Interesting task! I learned a lot of new things :)

Issue #92, opened 2024-03-21
Use more `role` in KeyUX


ai:
`:focus` navigation with only Tab can be not very useful because you will have a too long list of options.

We are using [KeyUX](https://github.com/ai/keyux) to replace Tab to arrows in some blocks like app menu or search results.

But the current UI code was created for KeyUX 0.4 and use only `role=menuitem` which is not fit all cases. New KeyUX 0.5 supports more `role` with arrow-navigation.

1. Read about `role` with focus grouping from [KeyUX docs](https://github.com/ai/keyux?tab=readme-ov-file#focus-groups). Check MDN or Doka docs for each role.
2. Find all `menuitem` usage in the web client.
3. Think where to replace `menuitem`.

dancetrain:
I want to take it

PR #131, opened 2024-03-30
Improve role attribute usage with KeyUX library

dancetrain:
Fixes [#92](https://github.com/hplush/slowreader/issues/92)

Changed role="menu" to role="menubar"
Changed links on Add Feed page to be an `option` of `listbox` role

## Motivation

Improved semantics of links/buttons for their purpose

## Checklist

- [x] Think about testing
- If you changed web client:
- If you changed `core/`:

### DIFF
--- a/web/ui/card-link.svelte
+++ b/web/ui/card-link.svelte
@@ -1,14 +1,16 @@
 <script lang="ts">
-  import { mdiChevronRight } from '@mdi/js'
+  import { mdiChevronRight, mdiListBox } from '@mdi/js'
   import { createEventDispatcher } from 'svelte'
 
   import Icon from './icon.svelte'
+  import type { AriaRole } from 'svelte/elements'
 
   export let href: string | undefined = undefined
   export let controls: string | undefined = undefined
   export let name: string
   export let current = false
   export let first = false
+  export let role: AriaRole = 'option'
 
   let dispatch = createEventDispatcher<{ click: null }>()
 
@@ -24,7 +26,7 @@
       aria-controls={controls}
       aria-current={current ? 'page' : null}
       {href}
-      role="option"
+      {role}
       tabindex={current || first ? null : -1}
       on:click={onClick}
     >
@@ -38,7 +40,7 @@
       class="card-link"
       aria-controls={controls}
       aria-current={current ? 'page' : null}
-      role="option"
+      {role}
       tabindex={current || first ? null : -1}
       on:click={onClick}
     >
--- a/web/ui/card-link.svelte
+++ b/web/ui/card-link.svelte
@@ -17,14 +17,14 @@
   }
 </script>
 
-<li role="presentation">
+<li role="none">
   {#if href}
     <a
       class="card-link"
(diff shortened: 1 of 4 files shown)


web/ui/card-link.svelte
```diff
@@ -17,14 +17,14 @@
   }
 </script>
 
-<li role="presentation">
+<li role="none">
```
ai, line 20: Why do we have this change?

web/ui/card-links.svelte
```diff
@@ -1,9 +1,12 @@
 <script lang="ts">
+  import type { AriaRole } from 'svelte/elements'
+
   export let node: HTMLUListElement | null = null
   export let id: string | undefined = undefined
+  export let role: AriaRole | undefined = 'listbox'
```
ai, line 6: Why we have a way to change role in CardLinks, but not in CardLink?

web/ui/navbar/index.svelte
```diff
@@ -104,7 +104,7 @@
     id="navbar_submenu"
     class="navbar_submenu"
     aria-hidden="true"
-    role="menu"
```
ai, line 107: Why we have menubar here if the submenu list is vertical?

web/ui/navbar/item.svelte
```diff
@@ -12,6 +13,7 @@
   export let secondary = false
   export let submenu = false
   export let small = false
+  export let role: AriaRole = 'menuitem'
```
ai, line 16: Where we plan to use this property?

I think adding options which we do not need make code readability worse. In my experience it is better to add option instead of inlined value only if we have plans to use option.

web/ui/card-link.svelte
dancetrain: - "presentation" and "none" are synonyms
- "none" is shorter

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/none_role

To be more representative - I think role="none" is more useful since from the naming you didn't expect it will be somehow iterate with UI/UX, while naming presentation - make give you false expectations

ai: Good idea

web/ui/navbar/item.svelte
dancetrain: I think during further re-factoring of role option we will need to pass sub role values to container element - I do not see other option to do that besides that to move it to properties

web/ui/navbar/index.svelte
dancetrain: Probably my bad, going to revert it, thank for gotcha this

web/ui/navbar/item.svelte
ai: What role do you plan to use in the future for the menu items?

ai:
If you do not have time for it anymore, we can move the issue back to `Onboarding`

dancetrain:
Sorry for delay, I wil continue work on this task

ai:
I moved issue back to Onboarding. But you can continue working on the PR.

PR #134, opened 2024-04-01
Improve proxy security (#100)

### LINKED ISSUE #100: Improve proxy security
We have HTTP proxy in `proxy/` folder to bypass CORS limits in the browser.

But I’m afraid that spammer could use it. Unfortunately, we can’t use authentication since we care about privacy.

- [x] Let’s allow only `GET` request for now, since `POST`/`PUT` is methods to change documents, and we don’t need them to load RSS.
- [x] Remove all cookie headers. Otherwise, our proxy will copy headers and set cookie for own domain. For instance, `cookie` from client and `set-cookie` from server.
- [x] In `process.env.NODE_ENV === 'production'` allow requests only from `*.slowreader.app` clients using `Origin` header.
- [x] Do not allow request to localhost IP addressed like `127.*` or `192.168.*` (check all possible local IP addresses), since it can be used to access to cloud infrastructure.
- [x] Remove `Host` header from client since it will be the proxy domain.
- [x] Add unit tests to proxy like we have in the core.

toplenboren:
Fixes #100 

I added some small improvements to `proxy.ts` service

I've implemented:
- Do not allow POST / PUT and pretty much everything besides GET
- Do not allow requests to reserved ips
- Do not allow anything besides http or https
- Remove shady headers
- Added a handy README note
- Test that init real target http server and proxy and test end to end user 

I asked a few questions in #100. Let's use issue to discuss planned changes

## Motivation

We do not want our proxy to be abused and banned, thus all the security stuff. Check #100 for details

## Deps

I needed a package to check whether IP is reserved or not. Thought that we do not want to bloat our code with this stuff. I found three packages: isMartianPacket, bogon and isReservedIp

I found [is-reserved-ip](https://www.npmjs.com/package/reserved-ip) code to be strange and [bogon](https://www.npmjs.com/package/bogon) to have one more dependency 👍 

I decided to pick [isMartianPacket](https://www.npmjs.com/package/martian-cidr).Package seems to be [maintained](https://github.com/arty-name/martian-cidr/issues/50), and I found the lack of types (Can implement this myself 🥨) and lack of adoption as main drawback (thou need to say that many people just introduce code for this task in their projects).  

## Checklist (Gonna be checked once we are out of draft state)

- [x] Run `pnpm test`.
- [x] Think about testing
- If you changed web client:
- If you changed `core/`:


proxy/README.md
```diff
@@ -3,3 +3,24 @@
 HTTP-server to proxy all RSS fetching request from web client.
 
 User could use it to bypass censorship or to try web client before they install upcoming extension (to bypass CORS limit of web app).
+
+## Usage
+
+### Start proxy
+
+```shell
```
ai, line 11:
```suggestion
```sh
```

proxy/README.md
```diff
@@ -3,3 +3,24 @@
...
+
+```sh
+pnpm start
+# // Proxy server running on port 5284
+```
+
+### Make a request
+
+```shell
+curl localhost:5284/https://hplush.dev
```
ai, line 19: Let’s not document the API 😏

proxy/package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "start": "tsx ./index.ts"
+    "start": "tsx ./proxy.ts"
```
ai, line 6: In all another project the main file is `index.ts` (like in `server/`). Let’s keep name system.

toplenboren: Makes sense :-) let's keep it systematical

proxy/README.md
toplenboren: Alright :-)

proxy/package.json
toplenboren: Done!

proxy/README.md
toplenboren: Done!

proxy/index.ts
```diff
@@ -1,43 +1,19 @@
...
-      }
-    } else if (typeof e === 'string') {
-      process.stderr.write(styleText('red', e) + '\n')
-    }
-  }
+const IS_PRODUCTION = process.env.mode === 'production'
+const PRODUCTION_DOMAIN_SUFFIX = '.slowreader.app'
+
+const proxy = createProxyServer({
+  isProduction: IS_PRODUCTION,
```
ai, line 11: Do we need constant if we use it right here?

proxy/proxy.ts
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * productionDomainSuffix - if isProduction, then only request from origins that match this param are allowed
+ *
+ * @example
+ * const p = createProxyServer()
+ * p.listen(5555).then(() => console.log('running on 5555'))
+ *
+ * @param { isProduction, silentMode, hostnameWhitelist, productionDomainSuffix } config
+ */
+const createProxyServer = (
```
ai, line 25: We do not use default export in the project (they sometimes have a problem in bundlers/tools, not a big dead but nice for code style—do we have ESLint rule for that?).

Let’s use `export function createProxyServer`

proxy/proxy.ts
```diff
@@ -0,0 +1,119 @@
...
+  let productionDomainSuffix = config.productionDomainSuffix || ''
+
+  return createServer(async (req, res) => {
+    let url = decodeURIComponent(req.url!.slice(1))
+    let sent = false
+
+    try {
+      // Only http or https protocols are allowed
+      if (!url.startsWith('http://') && !url.startsWith('https://')) {
+        throw new Error('Only http or https requests are allowed.')
```
ai, line 45: It will be nice to have `400` (Bad request) response here. It seems for me that `Error` is always will return `500`,

proxy/proxy.ts
```diff
@@ -0,0 +1,119 @@
...
+        'Content-Type':
+          targetResponse.headers.get('content-type') ?? 'text/plain'
+      })
+      sent = true
+      res.write(await targetResponse.text())
+      res.end()
+    } catch (e) {
+      if (e instanceof Error) {
+        if (!silentMode) {
+          process.stderr.write(styleText('red', e.stack ?? e.message) + '\n')
```
ai, line 100: We should not write `400` errors and timeout errors to the log, but need to write all other errors.

proxy/proxy.ts
```diff
@@ -0,0 +1,119 @@
...
+        }
+        if (!sent) {
+          res.writeHead(500, { 'Content-Type': 'text/plain' })
+          res.end('Internal Server Error')
+        }
+      } else if (typeof e === 'string') {
+        if (!silentMode) {
```
ai, line 107: I think here it will be better to re-use `if (!sent)` branch (by moving it out of the `if (typeof e === 'string') {`

proxy/test/proxy.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        queryParams,
+        requestPath,
+        url: req.url
+      },
+      response: 'ok'
+    })
+  )
+})
+
+describe('proxy tests', async () => {
```
ai, line 38: We don’t use single `describe` per file in other tests.

proxy/test/proxy.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    let target = getTestHttpServer('target')
+    if (target) {
+      targetServerUrl = target.baseUrl
+    }
+
+    if (!target || !proxy) {
+      throw new Error(
+        "Couldn't set up target server or proxy server. Something is misconfigured. Please check out 'proxy.test.js'"
```
ai, line 61: Can we make it shorter to fit `80` columns line?

proxy/test/proxy.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      )
+    }
+  })
+
+  await test('proxy works', async () => {
+    let response = await fetch(`${proxyServerUrl}/${targetServerUrl}`)
+    let parsedResponse = await response.json()
+    // @ts-ignore
```
ai, line 69:
```suggestion
    // @ts-expect-error
```

BTW, why?

ai: Maybe `equal(response?.status, 200)`

proxy/test/proxy.test.ts
```diff
@@ -0,0 +1,163 @@
...
+          let response = await fetch(`${proxyServerUrl}/${targetServerUrl}`, {
+            method
+          })
+          equal(response.status, 500)
+        })
+      }
+    })
+
+    await test('can use only http or https protocols', async () => {
+      let forbiddenProtocols = [
```
ai, line 102: A single test is enough :D

proxy/test/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+ * @param name
+ * @param server
+ * @param opts
+ */
+export async function initTestHttpServer(
+  name: string,
+  server: any,
+  opts?: { port?: number; protocol?: string }
+): Promise<void> {
+  //
```
ai, line 38:
```suggestion
```

proxy/test/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+ *   })
+ * })
+ *
+ * @param name
+ * @param server
+ * @param opts
+ */
+export async function initTestHttpServer(
+  name: string,
+  server: any,
```
ai, line 35: Not cool :)

proxy/test/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+  baseUrl: string
+  port: number
+  server: http.Server
+}
+
+const __testServers: {
+  [key: string]: TestServer
+} = {}
+
+/**
```
ai, line 15: We do not need JSDoc here. It is better to add it everywhere or not use it.

proxy/test/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+import { after, before } from 'node:test'
+
+interface TestServer {
+  address: string
+  baseUrl: string
+  port: number
+  server: http.Server
+}
+
+const __testServers: {
```
ai, line 11: Why not just `testServers`?

proxy/test/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+  opts?: { port?: number; protocol?: string }
+): Promise<void> {
+  //
+  before(async () => {
+    let port = opts?.port || 0
+    let protocol = opts?.protocol || 'http'
+
+    // port=0 is <any free port>
+    let testServerInstance = await server.listen(port)
+    // @ts-ignore
```
ai, line 45: What is the error here?

proxy/README.md
```diff
@@ -3,3 +3,18 @@
...
+## Usage
+
+### Start Proxy
+
+```sh
+pnpm start
+# // Proxy server running on port 5284
+```
+
+## Additional Info:
```
ai, line 16: Title should not contain `:`

proxy/proxy.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      res.writeHead(targetResponse.status, {
+        'Access-Control-Allow-Headers': '*',
+        'Access-Control-Allow-Methods': 'OPTIONS, POST, GET, PUT, DELETE',
+        'Access-Control-Allow-Origin': '*',
+        'Content-Type':
+          targetResponse.headers.get('content-type') ?? 'text/plain'
+      })
+      sent = true
+      res.write(await targetResponse.text())
```
ai, line 95: Do you want to also improve a performance a little?

Why do we need to first create a full copy of the HTTP document in the proxy memory and only then send it to the user? Instead, on receiving first bites we can send them to the client. It improves performance and memory consumption.

https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams

proxy/proxy.ts
```diff
@@ -0,0 +1,119 @@
...
+          ...(req.headers as HeadersInit),
+          host: new URL(url).host
+        },
+        method: req.method
+      })
+
+      res.writeHead(targetResponse.status, {
+        'Access-Control-Allow-Headers': '*',
+        'Access-Control-Allow-Methods': 'OPTIONS, POST, GET, PUT, DELETE',
+        'Access-Control-Allow-Origin': '*',
```
ai, line 90: There is a cool idea of using not `*` but `req.headers['Origin']`

proxy/proxy.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      // Remove all cookie headers and host header from request
+      delete req.headers.cookie
+      delete req.headers['set-cookie']
+      delete req.headers.host
+
+      let targetResponse = await fetch(url, {
+        headers: {
+          ...(req.headers as HeadersInit),
+          host: new URL(url).host
```
ai, line 82: Should we write client’s IP to avoid using proxy to hide IP? There is `X-Forwarded-For` header for that.

Later we will add privacy mechanism for paid users.

proxy/proxy.ts
```diff
@@ -0,0 +1,119 @@
...
+      res.writeHead(targetResponse.status, {
+        'Access-Control-Allow-Headers': '*',
+        'Access-Control-Allow-Methods': 'OPTIONS, POST, GET, PUT, DELETE',
+        'Access-Control-Allow-Origin': '*',
+        'Content-Type':
+          targetResponse.headers.get('content-type') ?? 'text/plain'
+      })
+      sent = true
+      res.write(await targetResponse.text())
+      res.end()
```
ai, line 96: There is another nice abuse protection system. We can limit the size of response https://github.com/gridaco/cors.sh/blob/main/services/proxy.cors.sh/src/limit/payload-limit.ts

proxy/index.ts
toplenboren: I just really like to put it on the top of the file in case we want to add additional logic on top of our proxy launch. 

By the way, what do you think about the way to configure proxy? 

We'd need to set process.env.mode to `production` on production.. How do you deploy this app? Is this kind of configuration ok for you?

proxy/proxy.ts
toplenboren: Fixed, thanks!

proxy/test/utils.ts
toplenboren: Just fixed that!

proxy/test/utils.ts
toplenboren: Just checked out other files and added a small commentary on the top of module instead of docstrings

proxy/test/utils.ts
toplenboren: I usually like to indicate that this variable is private, but let's be simple here! Changed back to `testServers`

proxy/README.md
```diff
@@ -3,3 +3,18 @@
 HTTP-server to proxy all RSS fetching request from web client.
 
 User could use it to bypass censorship or to try web client before they install upcoming extension (to bypass CORS limit of web app).
+
+## Usage
```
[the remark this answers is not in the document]
toplenboren: Fixed `README.md`

proxy/test/utils.ts
toplenboren: Just fixed that

proxy/test/utils.ts
toplenboren: The error was in the way `.addressInfo()` data was shaped. I think I figured it out with type conversion

proxy/README.md
toplenboren: Just fixed `README.md`

proxy/test/proxy.test.ts
toplenboren: Can we have a single describe here? The reason is that I want to utilise `before` and `after` so that I can init and destroy http server. 

If I ditch single describe, then I'd have to write a lot of:

```js
await initTestHttpServer(
    'proxy',
    createProxyServer({ hostnameWhitelist: ['localhost'], silentMode: true })
  )
  await initTestHttpServer('target', targetServer)

  let proxyServerUrl = ''
  let targetServerUrl = ''

  before(() => {
    let proxy = getTestHttpServer('proxy')
    if (proxy) {
      proxyServerUrl = proxy.baseUrl
    }

    let target = getTestHttpServer('target')
    if (target) {
      targetServerUrl = target.baseUrl
    }

    if (!target || !proxy) {
      throw new Error(
        "Couldn't set up target server or proxy server. Please check out 'proxy.test.js'"
      )
    }
  })
```

Maybe there is a way to support before outside the single describe?

proxy/test/proxy.test.ts
toplenboren: Just made that comment shorter! Thanks

proxy/test/proxy.test.ts
toplenboren: Just fixed this, thanks!

proxy/test/proxy.test.ts
toplenboren: I agree with you, but I think it is better to just add more protocols and don't think of it again :D

proxy/proxy.ts
toplenboren: Agreed. Code will smell better with this change! Just fixed that

proxy/proxy.ts
toplenboren: Seems cool. Just added this feature

proxy/proxy.ts
toplenboren: Agreed. Let me take some time to research usecases and write tests :D

proxy/proxy.ts
toplenboren: Agreed, added 400 response here

proxy/proxy.ts
toplenboren: Agreed, just fixed that

proxy/proxy.ts
toplenboren: Agreed, just fixed that

proxy/proxy.ts
toplenboren: Agreed, just fixed that

toplenboren:
Hey, @ai I've done some cool stuff! Can you please re-review my PR? 

## Done:

- Introduced timeout feature
- Refactored error handling: All Bad Requests will not be looged and will return 400 to user
- Fixed minor issues, such as README.md structure or typescript errors

## Questions:

- I have a question about single `describe` in tests. Please check it
- I have a question about configuration

## Left to do:

1. Debounce
2. Hide client IP
3. Limit size of response
4. Introduce partial return for optimizations

proxy/index.ts
ai: I will do it when I will prepare deploy (it need to be done together with Google Cloud access for tests).

For instance, we will nee to read port from `env.PORT`, etc.

proxy/proxy.ts
```diff
@@ -0,0 +1,126 @@
+// @ts-ignore
```
ai, line 1: Let’s define simple types in `global.d.ts`. It should be a few lines just for API which we use.

proxy/proxy.ts
```diff
@@ -0,0 +1,126 @@
...
+export const createProxyServer = (
+  config: {
+    fetchTimeout?: number
+    hostnameWhitelist?: string[]
+    isProduction?: boolean
+    productionDomainSuffix?: string
+    silentMode?: boolean
+  } = {}
+): http.Server => {
+  // Main toggle for production features
```
ai, line 24: The arguments are very simple to explain them in comments (in contrast with different hacks)

proxy/README.md
```diff
@@ -3,3 +3,23 @@
...
+
+```sh
+pnpm start
+# // Proxy server running on port 5284
+```
+
+## Abuse Protection
+
+- Proxy allows only GET requests
+- Proxy do not allow requests to reserved ip addresses like `127.0.0.1`
```
ai, line 19:
```suggestion
- Proxy do not allow requests to in-cloud IP addresses like `127.0.0.1`
```

proxy/README.md
```diff
@@ -3,3 +3,23 @@
...
+- Proxy allows only http or https
```
ai, line 20:
```suggestion
- Proxy allows only HTTP or HTTPS protocols
```

proxy/README.md
```diff
@@ -3,3 +3,23 @@
...
+- Proxy removes cookie headers
+
+## Test Strategy
+
+Proxy is tested using e2e testing. To write tests `initTestHttpServer` should be used
```
ai, line 25:
```suggestion
To test proxy we emulate the real HTTP servers (end-to-end testing).
```

Docs are for juniors.

proxy/proxy.ts
```diff
@@ -0,0 +1,126 @@
...
+import { styleText } from 'node:util'
+
+class BadRequestError extends Error {
+  constructor(message: string) {
+    super(message)
+    this.name = 'BadRequestError'
+  }
+}
+
+export const createProxyServer = (
```
ai, line 15: Let’s use `function` if it is a function

proxy/proxy.ts
```diff
@@ -0,0 +1,126 @@
...
+      if (isProduction) {
+        let origin = req.headers.origin
+        if (!origin?.endsWith(productionDomainSuffix)) {
+          throw new BadRequestError('Unauthorized Origin')
+        }
+      }
+
+      let requestUrl = new URL(url)
+      if (!hostnameWhitelist.includes(requestUrl.hostname)) {
+        // Do not allow requests to addresses that are reserved:
```
ai, line 59:
```suggestion
        // Do not allow requests to addresses inside our cloud:
```

We should explain “why we did it” not “what we did”

proxy/proxy.ts
```diff
@@ -0,0 +1,126 @@
...
+        if (
+          (isIP(requestUrl.hostname) === 4 ||
+            isIP(requestUrl.hostname) === 6) &&
+          isMartianIP(requestUrl.hostname)
+        ) {
+          throw new BadRequestError('Requests to reserved ips are not allowed')
+        }
+      }
+
+      // Remove all cookie headers and host header from request
```
ai, line 72:
```suggestion
      // Remove all cookie headers so they will not be set on proxy domain
      // Replace Host header to target
```

proxy/test/proxy.test.ts
ai: You can use `before`/`after` without `description`.

Just use `beforeEach` or `beforeAll` from `node:test`.

proxy/test/utils.ts
```diff
@@ -0,0 +1,68 @@
+//  Use this function if you need to init local http server for testing. Express is supported. Server will be closed automatically after tests finish
```
ai, line 1: We don’t need docs here since the test is a good example

PR #135, opened 2024-04-02
[Add]: formatted text styles

### LINKED ISSUE #118: Define styles for all post’s tag
We have XSS protection for post’s body and define allow-list of [all possible tags](https://github.com/hplush/slowreader/blob/main/core/html.ts#L3) in post.

We need to add base styles for that tags and check them in Storybook.

- [ ] Add styles to `web/ui/formatted-text.svelte`.
- [ ] Add `web/ui/stories/ui/formatted-text.stories.svelte` story with all possible cases. Try to use the single `<Story>` (we are on the free plan in Chromatic with snapshots limits 🥲)

upteran:
Fixes #118

- Add base styles for allowed tags from the [allowed list](https://github.com/hplush/slowreader/blob/main/core/html.ts#L3) in FormattedText component
- Add a story for the FormatterText component with different HTML tag examples
- Add CSS custom properties for a common list for link and code-block tags.

## Motivation

The changes were made to the styles in the formatter text component to enhance the visual appeal and readability of the long heading text. 

## Screenshots

Few screenshots to show some frequently used tags. The entire list of other tags has been added to the story.

Screenshot 2024-04-02 at 10 29 00
Screenshot 2024-04-02 at 10 29 53
Screenshot 2024-04-02 at 10 30 58
Screenshot 2024-04-02 at 10 31 34

## Checklist

  - [ ] Look for alternatives, not just take the most popular one.
  - [ ] Check alternatives for [bundle size](https://bundlejs.com/), [`node_modules` size](https://packagephobia.com/), and [subdependencies number](https://npmgraph.js.org).
  - [ ] Check that the project is active by their repository/issues/PR.
  - [ ] Run `cd web && pnpm size`, check the JS bundle size changes, about limits if necessary.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:

## Questions during PR:

I have few questions about PR:

- I added the `createTable` function to build table examples, but I'm not sure if it's necessary because it's probably complicated code a simple few row/cell table by string might be enough. 
- Regarding image links, I want to add some to the stories as examples, but I'm not sure if external links to other sites are a good choice.
- Added some custom variables, but perhaps they should be placed in the FormatterText component? However, in this case, we might forget about these variables if we change some styles for the theme.
- And about the story, I've seen separate stories for light and dark mode in other components (number 2 in the screenshot), but without this separation, I can still check the story without creating two different ones by switching the theme toggler (number 1 in the screenshot). Do we need to add two components with different theme prop configurations, or is the Base enough?
Screenshot 2024-04-02 at 10 40 43 1


ai:
> Do we need to add two components with different theme prop configurations, or is the Base enough?

Yes, we need separated stories.

The problem with theme switcher is that we can’t use it on Chromatic (screenshot based testing).

Another reason is that now checking Storybook by eyes is very straight-forward—you just going from one story to another and check it by eyes. With theme switcher you need to remember to manually test every story in different themes, which is more complex than clicking `Next Story`.

web/main/colors.css
```diff
@@ -6,6 +6,11 @@
   --fire1-color: oklch(70% 0.15 65);
   --fire2-color: oklch(58% 0.2 29);
   --error-color: var(--red-600);
+  --code-inline-bg-color: color-mix(in oklch, var(--neutral-700), var(--neutral-600));
+  --code-inline-color: oklch(100% 0 0);
+  --link-color-hover: oklch(from var(--link-color) l c h / 85%);
+  --link-color-visited: oklch(from var(--link-color) l c h / 75%);
```
ai, line 12: I'm afraid that 75% is not enough for good readability.

Also, the classic blue→violet color changes is a more standard way.

web/main/colors.css
```diff
@@ -6,6 +6,11 @@
   --fire1-color: oklch(70% 0.15 65);
   --fire2-color: oklch(58% 0.2 29);
   --error-color: var(--red-600);
+  --code-inline-bg-color: color-mix(in oklch, var(--neutral-700), var(--neutral-600));
+  --code-inline-color: oklch(100% 0 0);
+  --link-color-hover: oklch(from var(--link-color) l c h / 85%);
```
ai, line 11: Can we use `opactiy`?

web/main/colors.css
```diff
@@ -45,6 +50,7 @@
       0 -1px oklch(0% 0 0 / 20%);
     --field-color: oklch(100% 0 0);
     --field-shadow: inset 0 1px 2px oklch(from var(--shadow-color) l c h / 8%);
+    --link-color: oklch(49.77% 0.224 271.1);
```
ai, line 53: It is possible to take color from Harmony?
https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://www.figma.com/community/file/1287828769207775946/harmony-accessible-ui-color-palette&ved=2ahUKEwjppZyO8qOFAxXQQUEAHewQDu0QFnoECBQQAQ&usg=AOvVaw0BjztSBE5b3U7HmRkyQdM3

web/ui/formatted-text.svelte
```diff
@@ -17,8 +17,200 @@
...
+
+  .formatted-text :global(p code) {
+    padding: 0.3em;
+    font-size: 80%;
+    color: var(--code-inline-color);
+    background: var(--code-inline-bg-color);
+    border-radius: var(--radius);
+  }
+
+  /* !** table *! */
```
ai, line 187: Let’s remove comments.

web/stories/ui/formatted-text.stories.svelte
```diff
@@ -0,0 +1,201 @@
...
+    '  <li><a href="https://example.com">Item with link</a></li>\n' +
+    '  <li><strong>Strong list item</strong></li>\n' +
+    '  <li>Item <i>with itallic formatt</i></li>\n' +
+    '</ul>'
+
+  let orderedList = unorderedList.replaceAll('ul', 'ol')
+
+  let contentTags =
+    '  <figure>\n' +
+    '    <img src="https://media.dev.to/cdn-cgi/image/width=400,height=220,fit=cover,gravity=auto,format=auto/https%3A%2F%2Fdev-to-uploads.s3.amazonaws.com%2Fuploads%2Farticles%2F7u4wo1pru5qle7jj67of.png" alt="Example image">\n' +
```
ai, line 58: You can use images from `web/public/`. It will be more reliable and works in offline.

web/ui/formatted-text.svelte
```diff
@@ -17,8 +17,200 @@
...
+    font-size: 0.89em;
+    background: transparent;
+  }
+
+  .formatted-text :global(p code) {
+    padding: 0.3em;
+    font-size: 80%;
+    color: var(--code-inline-color);
+    background: var(--code-inline-bg-color);
+    border-radius: var(--radius);
```
ai, line 184: This radius is too big for small elements. It is OK to avoid token here (or create another token).

web/ui/formatted-text.svelte
```diff
@@ -17,8 +17,200 @@
...
+    max-width: 100%;
+    padding: 0.1em 0.25em;
+    font-size: 0.89em;
+    background: transparent;
+  }
+
+  .formatted-text :global(p code) {
+    padding: 0.3em;
+    font-size: 80%;
+    color: var(--code-inline-color);
```
ai, line 182: The black code block is not look good. Can we try another way? GitHub has nice style for `code`.

web/stories/ui/formatted-text.stories.svelte
```diff
@@ -0,0 +1,201 @@
...
+      'Header 1',
+      'Header 2',
+      'Header 3',
+      'Header 4',
+      'Header 5',
+      'Header 6',
+      'Header 7'
+    ]
+  }
+  function createTable(h: string[], d: string[][]): string {
```
ai, line 127: Yes, maybe it is better to just put tags manually once.

web/main/colors.css
upteran: Replaced with violet color from Harmony

web/main/colors.css
upteran: Yes, it makes sense. It's definitely a better decision in this case. done

web/main/colors.css
upteran: Omg how I missed this tool, awesome, yes let's take color from Harmony. Should we use these colors in most cases? If so, maybe we should add some notes to the 'Visual Language' section in the documentation.

web/main/colors.css
```diff
@@ -45,6 +46,9 @@
       0 -1px oklch(0% 0 0 / 20%);
     --field-color: oklch(100% 0 0);
     --field-shadow: inset 0 1px 2px oklch(from var(--shadow-color) l c h / 8%);
+    --link-color: var(--blue-700);
+    --link-color-visited: var(--violet-700);
+    --code-inline-bg-color: oklch(from var(--neutral-200) l c h / 55%);
```
ai, line 51: We do not use `bg` word in other tokens. `--inline-code-color` will fit the system better.

web/stories/ui/formatted-text.stories.svelte
```diff
@@ -0,0 +1,227 @@
+<script context="module" lang="ts">
+  import imgExample2 from '../../public/stories-assets/images/long_width_example.avif'
```
ai, line 2: If we are adding new images, it is better to put them to `web/stories/ui/assets/`.

The best way to get images is Wikipedia since all of they are free (EM logo is not open source).

ai: Also, we need a way to avoid adding those images to Size Limit. For instance, we can use some prefix and add its ignore to `web/.size-limit.json`. Or we can configure Vite to put all `web/stories/` images to some subdir (and ignore it in the same `web/.size-limit.json`).

web/ui/formatted-text.svelte
```diff
@@ -20,5 +20,179 @@
...
+    margin: 1em 0;
+    font-size: 1.067em;
+  }
+
+  .formatted-text :global(a) {
+    color: var(--link-color);
+    text-decoration: none;
+    word-wrap: break-word;
+    opacity: 100%;
+    transition:
```
ai, line 66: Having `color` transition on `:hover` makes UI slow. since UI has no instant reaction of user’s actions.

Animations are cool, but we should use them to explain UI changes.

web/stories/ui/formatted-text.stories.svelte
```diff
@@ -0,0 +1,227 @@
...
+    '  <p>Did you know that the <data value="whale-shark">whale shark</data> is the largest fish in the world?</p>\n' +
+    '  <p><del>Conclusion</del></p>\n' +
+    '  <p><ins>Fish are fascinating creatures!</ins> They come in all shapes and sizes, from the tiny <small>paedocypris</small> to the massive <samp>whale shark</samp>.</p>\n' +
+    '  <p><sub>Subscript text</sub> and <sup>superscript text</sup> can also be used for scientific notation.</p>\n' +
+    '  <p>For more information, visit our <a href="https://example.com">website</a>.</p>\n'
+
+  let unorderedList =
+    '<ul>\n' +
+    '  <li>Simple list item</li>\n' +
+    '  <li><a href="https://example.com">Item with link</a></li>\n' +
```
ai, line 51: We need a visual test for new and visited link.

I suggest:

1. For new link use random domain `example + random + .com`
2. For visited link we can use `href=""` to the current page

web/main/colors.css
```diff
@@ -78,6 +82,9 @@
     --flat-active-shadow: inset 0 1px 2px oklch(0% 0 0 / 80%);
     --field-color: color-mix(in oklch, var(--neutral-900), var(--neutral-800));
     --field-shadow: inset 0 1px 2px oklch(0% 0 0 / 60%);
+    --link-color: var(--blue-300);
```
ai, line 85: Lighthouse complains about link color contrast:

```
Links rely on color to be distinguishable.
```

web/stories/ui/formatted-text.stories.svelte
upteran: I have thoughts about this case. Storybook builds independently, but assets from it now are copied to the `dist/assets` folder of our main project. However, if we place these assets, for example, in `web/stories/assets`, it will go to `dist/ui/assets`, to the stories folder. In this case, size-limit will skip the check (since it already ignores `!dist/ui/**`), and we'll divide the files of packages. I don't think we'll use these images in the main project ? What do you think ?

ai: Nice idea

web/stories/ui/formatted-text.stories.svelte
```diff
@@ -0,0 +1,228 @@
...
+    '  <p>Did you know that the <data value="whale-shark">whale shark</data> is the largest fish in the world?</p>\n' +
+    '  <p><del>Conclusion</del></p>\n' +
+    '  <p><ins>Fish are fascinating creatures!</ins> They come in all shapes and sizes, from the tiny <small>paedocypris</small> to the massive <samp>whale shark</samp>.</p>\n' +
+    '  <p><sub>Subscript text</sub> and <sup>superscript text</sup> can also be used for scientific notation.</p>\n' +
+    '  <p>For more information, visit our <a href="https://example.com">website</a>.</p>\n'
+
+  let unorderedList =
+    '<ul>\n' +
+    '  <li>Simple list item</li>\n' +
+    '  <li><a href="https://example + random + .com">Item with not visited link</a></li>\n' +
```
ai, line 51: You need to put a random number there :)

```suggestion
    `  <li><a href="https://example${Math.round(Math.random()*1000)}.com">Item with not visited link</a></li>\n` +
```

web/main/colors.css
upteran: Can I ask what page you tested? Because I tried checking `feeds/categories/` with Lighthouse and didn't receive this message. However, I did get it in the story, probably due to the storybook links. Some of the low contrast color buttons at the top right are links. Maybe it's problem that Lighthouse defines

web/stories/ui/formatted-text.stories.svelte
```diff
@@ -0,0 +1,228 @@
...
+    '  <li><a href="">Item with visited link</a></li>\n' +
```
ai, line 52: It is not marked as visited:

Captura desde 2024-04-04 23-57-40

https://preview-135---staging-3ryqvfpd5q-ew.a.run.app/ui/?path=/story/ui-formattedtext--base

web/main/colors.css
```diff
@@ -45,6 +45,9 @@
       0 -1px oklch(0% 0 0 / 20%);
     --field-color: oklch(100% 0 0);
     --field-shadow: inset 0 1px 2px oklch(from var(--shadow-color) l c h / 8%);
+    --link-color: var(--blue-700);
+    --link-color-visited: var(--violet-700);
```
ai, line 49: The color difference between visited and non-visited is too small

Captura desde 2024-04-04 23-57-47

web/main/global.d.ts
```diff
@@ -17,3 +17,6 @@ interface Navigator {
       }
     | undefined
 }
+
+declare module '*.avif'
```
ai, line 21: Can I ask you to find smaller images on Wikipedia?

300 KB is too much.

For PNG, it is better to use some logotype, because photo is bad in PNG compression.

Also, you can use https://squoosh.app/ to convert it to AVIF or WebP to reduce the size.

I had a few projects where tests assets after 2 years of project history bloat `.git` size.

web/stories/ui/formatted-text.stories.svelte
upteran: ok i'll check, I have different colors in my browser

ai: I use Firefox to test

web/stories/ui/formatted-text.stories.svelte
upteran: aaw, I get, I thought it was some kind of hack that I didn't know, because it worked) ok of course will use function

web/main/colors.css
upteran: maybe use this one, probably looks good 
Screenshot 2024-04-05 at 00 47 00

ai: Yes, perfect

PR #144, opened 2024-04-03
Mobile navbar

### LINKED ISSUE #94: Add mobile version of navbar
We need to add mobile layout to the web client. In this task we will start from `<Navbar>`.

- [ ] By default, hide feeds/settings list. Make menu just like a line.
Captura desde 2024-03-22 01-45-17
- [ ] Move it to the bottom of the screen for mobile and tablets
- [ ] Make it float over content. Be sure that the page has padding on the bottom, so bottom page’s content will not be hidden by floating menu.
- [ ] Show feeds/settings list above the menu when user click on Slow/Fast/Menu items.
- [ ] Hide feeds/settings list when user selects an item there.
- [ ] Think about tablet menu design.
- [ ] Add mobile/tablet stories to the Storybook.
- [ ] Check that settings page looks good. Ignore pages with `<TwoStepsPage>` for now.

AnnaGolmakova:
Fixes #94 

Added mobile layout to the Navbar component for the web client:

- [x] Move it to the bottom of the screen for mobile and tablets
- [x] Make it float over content. Be sure that the page has padding on the bottom, so bottom page’s content will not be hidden by floating menu.
- [x] Show feeds/settings list above the menu when user click on Slow/Fast/Menu items.
- [x] Hide feeds/settings list when user selects an item there.
- [ ] Think about tablet menu design.
   * Used the same appearance as in mobile
- [ ] Add mobile/tablet stories to the Storybook.
   * Added a story showing navigation in a mobile viewport
   * Would it be enough?
- [x] Check that settings page looks good. Ignore pages with <TwoStepsPage> for now.

## Checklist

- [x] Run `pnpm test`.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:


pnpm-lock.yaml
```diff
@@ -241,6 +241,9 @@ importers:
       '@storybook/addon-themes':
```
ai, line 241: Call `pnpm install` another time and push changes.

CI fails because lock file is old, it happens often.

web/main/index.css
```diff
@@ -47,4 +47,8 @@ ol[role='list'] {
   box-sizing: border-box;
   width: 100%;
   padding-inline-start: var(--navbar-width);
+
+  @media (width <=1024px) {
```
ai, line 51:
```suggestion
  @media (width <= 1024px) {
```

web/ui/navbar/index.svelte
```diff
@@ -134,13 +149,25 @@
     display: flex;
     flex-direction: column;
     width: var(--navbar-width);
+
+    @media (width <=1024px) {
+      inset-block: unset;
+      bottom: 0;
+      width: 100%;
+      background-color: var(--land-color);
+      box-shadow: 0 -1px 0 oklch(from var(--shadow-color) l c h / 8%);
```
ai, line 158: You need to create a new token if you want another shadow

web/ui/navbar/other.svelte
```diff
@@ -22,6 +22,7 @@
   href={getURL('add')}
   icon={mdiPlusCircleOutline}
   secondary
+  on:click
```
ai, line 25: Tell me more how does `on:click` works here.

I am not very advanced Svelte developer too.

web/stories/ui/navbar.stories.svelte
```diff
@@ -51,6 +51,17 @@
   </Scene>
 </Story>
 
+<Story
```
ai, line 54: We may need also a story for small tablets. How menu will look like on 800px?

web/main/index.css
AnnaGolmakova: It seems there is a rule somewhere which prevents me from adding this space in .css files. I haven't figured out where it is yet. I fixed this in Svelte files.

ai: Does it remove space on saving the file or during making the commit?

web/main/colors.css
```diff
@@ -24,30 +24,28 @@
       inset 2px 0 0 var(--land-color),
       5px 0 0 var(--land-color),
       -5px 0 0 var(--land-color);
+    --navbar-shadow: 0 -1px 0 oklch(from var(--shadow-color) l c h / 8%);
```
ai, line 27: What do you think if we made token more universal?

It is not just for navbar only but for any floating component. Something like `--float-shadow`.

web/stories/ui/navbar.stories.svelte
AnnaGolmakova: Should I make a story for tablets even if the design for tablets is the same as for mobile? Or it would be enough that inside the mobile story you can change a viewport?

web/main/index.css
AnnaGolmakova: It removes space when I try to save the file.

web/main/colors.css
AnnaGolmakova: It's a good idea! I've changed the name

web/stories/ui/navbar.stories.svelte
ai: > Should I make a story for tablets even if the design for tablets is the same as for mobile?

Yes. Because we need to check how this design will scale to tablet.

For instance, we can add spacers around Fast/Slow or stretch those buttons (I am not sure that it will look good).

And we need to be sure that we will not have regression in design (broken spacers).

Also, later we will have a little changes for tablet UI.

web/main/index.css
ai: It means that it is something in IDE.

My suggestion:
1. Disable Prettier restart IDE and try again.
2. Try to look plugins for something relevant and try to disable it (and restart to check).
3. Disable all plugins.

ai:
My design suggestions:

1. Make shadow a little bigger.
2. Let’s make shadow mush darker for dark theme. Otherwise, it will be less visible.
3. Can you remove `cursor: default` on mobile/tablet for `.navbar-item[aria-current="page"]`? We have it because there is no need to click on active section again. But on mobile you may need to click it to see menu.
4. Very long Fast/Slow buttons on tablet doesn’t look very good. What do you think about some spacers and max width?

web/stories/ui/navbar.stories.svelte
AnnaGolmakova: For now I've added a story for a tablet but I haven't experimented with design yet.

ai: This is a good thing about stories (separated stories for each case, and not changing viewport in Storybook).

It forces thinking about edge cases.

web/main/index.css
AnnaGolmakova: I've corrected it manually

AnnaGolmakova:
> 4. Very long Fast/Slow buttons on tablet doesn’t look very good. What do you think about some spacers and max width?

Ok, I'll think about it and try to figure something out

ai:
Do you plan to continue? If not, I can finish it.

Or if you need some advice, I can help.

AnnaGolmakova:
> Do you plan to continue? If not, I can finish it.
> 
> Or if you need some advice, I can help.

Yes, I would like to finish it. Sorry for the delay

web/ui/navbar/other.svelte
AnnaGolmakova: In Svelte component events don't bubble. So, if you want to listen to an event on a deeply nested component, the intermediate components need to forward the event.

AnnaGolmakova:
> My design suggestions:
> 
> 1. Make shadow a little bigger.
> 2. Let’s make shadow mush darker for dark theme. Otherwise, it will be less visible.
> 3. Can you remove `cursor: default` on mobile/tablet for `.navbar-item[aria-current="page"]`? We have it because there is no need to click on active section again. But on mobile you may need to click it to see menu.
> 4. Very long Fast/Slow buttons on tablet doesn’t look very good. What do you think about some spacers and max width?

I've corrected these

web/ui/navbar/other.svelte
ai: Nice to know, thanks

ai:
Thanks! Looks amazing.

PR #147, opened 2024-04-04
Replace picocolors imports from scripts, remove picocolors dependency…

### LINKED ISSUE #117: Replace `picocolors` to `util.styleText`
Upcoming Node.js 20.12 will have built-in function for color output: [`util.styleText`](https://nodejs.org/en/blog/release/v20.12.0#text-styling)

1. Replace `picocolors` in our scripts, etc.
2. Remove `picocolors` dependency.

Edward-Leopold:
Fixes #117 

- replace picocolors imports from scripts
- replace pico functions with styleText func from node:util module
- remove picocolors dependency
- add global.d.ts file at the root of the project

## Motivation

I've added global.d.ts file because, probably, the @types/node have not updated yet, which is why the upcoming function styleText isn't available

## Screenshot or video
image

## Checklist

- [ ] Run `pnpm test`.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -1,3 +1,5 @@
-declare module 'node:utils' {
+declare module 'node:util' {
+  // Hack until this PR will be merged
+  // https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69212
   export function styleText(format: string, text: string): string
 }
--- a/global.d.ts
+++ b/global.d.ts
@@ -1,0 +1,3 @@
+declare module 'node:utils' {
+  export function styleText(format: string, text: string): string
+}
--- a/loader-tests/package.json
+++ b/loader-tests/package.json
@@ -9,7 +9,6 @@
     "@slowreader/core": "link:../core",
     "jsdom": "24.0.0",
     "nanostores": "0.10.2",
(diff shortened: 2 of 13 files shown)


ai:
Seems like you forgot to push `globals.d.ts` from root which we talked about in Telegram

global.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module 'node:utils' {
```
ai, line 1: You have `utilS` here but `util` in the imports

global.d.ts
```diff
@@ -0,0 +1,3 @@
...
+  export function styleText(format: string, text: string): string
```
ai, line 2: Let’s also add comment that it is a temporary hack. Something like:

```js
  // Hack until this PR will be merged
  // https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69212
  export function styleText(format: string, text: string): string
```

Those comments help to clean hacks when you don’t need them anymore. Without them, it is easy to find yourself year later with hack 100+ hacks when you don’t know what you can delete now.

ai:
Thanks! Congrats with first PR to the project.

PR #149, opened 2024-04-05
Add more Markdown plugins

### LINKED ISSUE #140: Add more Markdown plugins
The second step of #97 task after #129.

We need to add a few custom Remark plugins.

- [x] That TOC has right titles and URLs.
- [x] That titles have the correct word's capitalization (a/the is small, noun is starting from big latter).
- [ ] That file use curvy typography "→”, '→’.
- [x] That code block inside list do not split the list (has the right indent)

We can release those Remark plugins as open source project (or do it as internal plugins).

ilyatitovich:
Fixes #140 

I started working on custom plugins for `remark`. The first plugin I decided to make is a plugin for checking the correct capitalization of words in titles.

Additional packages have been installed: `unified-lint-rule` and `unist-util-visit`. This packages allow us to create custom rules for `remark`.

#### What does the plugin do?
Plugin formats the title, compares it with the current one and issues a warning with the expected variant and what was found.

I decided to do full formatting of the title so that the plugin shouldn't skip some strange mistakes, for example `ExAmplE woRd`. If user needs to leave the spelling of a word, he can add it to the exclusion list.

The formated title is saved in cache, which allow us to avoid repeated calculations if the same title is found in other documents.

#### Logic
To detect articles and other words that, according to the rules, should be written with a small letter, I created an object where these words will be stored. This will allow us to quickly find the desired exception by checking the key.

I added the ability to send to the plugin a list of exception words that the linter should skip. This will allow us to bypass the rules if the author believes that the word should retain its spelling.

The linter also skip words written in uppercase. It is understood that such writing is the original intention of the author.

Linter will detect if the first word in the title is capitalized (even if it is usually written in lowercase) and will suggest capitalizing it.

The result of the plugin's work currently looks like this:
titles

This is the first project of this kind for me, I’m looking forward to the review! :)

## Motivation

Our own plugins will allow us to find specific errors in `.md` files.

## Checklist

- [x] Run `pnpm test`.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:


.remarkrc.js
```diff
@@ -1,5 +1,14 @@
 // Configuration file used to find popular mistakes in .md files.
 
+import remarkLintTitleCapitalization from './remark-custom-rules/title-capitalization/remarkLintTitleCapitalization.js'
```
ai, line 3: Let’s use `remark/` folder, as we have `postcss/`

.remarkrc.js
```diff
@@ -1,5 +1,14 @@
...
+
 export default {
-  plugins: ['remark-lint-no-dead-urls', 'remark-validate-links']
+  plugins: [
+    'remark-lint-no-dead-urls',
+    'remark-validate-links',
+    [
+      remarkLintTitleCapitalization,
+      { exceptionWords: { '2FA': true, 'GitHub': true } }
```
ai, line 11: The wrong upper case letter inside the word is very rare use case.

What if will not check it? Only the first letter of the word?

remark-custom-rules/title-capitalization/remarkLintTitleCapitalization.js
```diff
@@ -0,0 +1,75 @@
...
+
+    // If the title is found among the correct titles - no calculations are performed.
+    if (cache[title]) {
+      return
+    }
+
+    const titleCheckResult = checkTitle(title, options.exceptionWords)
+
+    if (!titleCheckResult.isCorrect) {
+      file.message(
```
ai, line 62: ESLint and Stylelint has auto-fixable rules, which report to linter what is expected node and linter can fix document automatically.

Maybe Remark support it as well?

remark-custom-rules/title-capitalization/remarkLintTitleCapitalization.js
```diff
@@ -0,0 +1,75 @@
...
+      }
+
+      return word
+    })
+    .join(' ')
+
+  // Putting correct title in the cache for prevent handling the same titles in other docs.
+  cache[correctTitle] = correctTitle
+
+  const isCorrect = correctTitle === title
```
ai, line 45: This code is too complex.

I think this should be more easy to read:

```js
if (correctTitle === title) {
  return { isCorrect: true }
} else {
  return { isCorrect: false, correctTitle }
}
```

remark-custom-rules/title-capitalization/remarkLintTitleCapitalization.js
```diff
@@ -0,0 +1,75 @@
+/*
```
ai, line 1: In the project we use TS as much as possible.

But if you want to publish plugin as open source project, it is easier to publish JS projects than TS.

remark-custom-rules/title-capitalization/remarkLintTitleCapitalization.js
```diff
@@ -0,0 +1,75 @@
...
+
+  return isCorrect ? { isCorrect } : { isCorrect, correctTitle }
+}
+
+function titleCapitalization(tree, file, options = {}) {
+  visit(tree, 'heading', node => {
+    const title = node.children.map(child => child.value).join('')
+
+    // If the title is found among the correct titles - no calculations are performed.
+    if (cache[title]) {
```
ai, line 55: I like having the cache.

Also if you like micro-optimization, you can use faster logic for one-word titles.

remark-custom-rules/title-capitalization/remarkLintTitleCapitalization.js
```diff
@@ -0,0 +1,75 @@
...
+      return
+    }
+
+    const titleCheckResult = checkTitle(title, options.exceptionWords)
```
ai, line 59: I think you can make code simpler if you replace `checkTitle` to `fixTitle`.

Then you will have:

```js
let correctTitle = fixTitle(title, options.exceptionWords)
if (title !== correctTitle) {
  file.message(`Title's words capitalization. Expected: '${correctTitle}' found: '${title}'`, node)
}
```

Returning complex structure always slow down readability (it is good for special use cases, but here we can return just single value)

remark-custom-rules/title-capitalization/remarkLintTitleCapitalization.js
```diff
@@ -0,0 +1,75 @@
...
+
+          // Checking that word shouldn't be lowercase or if it is the first word in the title.
+        } else if (!lowerCaseWords[word] || index === 0) {
+          word = capitalizeWord(word)
+        }
+      }
+
+      return word
+    })
+    .join(' ')
```
ai, line 40: Not all spaces are equal :).

I often use ` ` (non-break space) to prohibit breaking line in some part of the sentence.

You can process words with saving spaces by:

```js
// Even Flight-or-Fight case is processed there
title.replace(/[^\s-]\w+/, word => {
  
})
```

ilyatitovich: Cool way to handle strings! The code has become much cleaner :)

remark-custom-rules/title-capitalization/remarkLintTitleCapitalization.js
ilyatitovich: I agree, it's much better this way

remark-custom-rules/title-capitalization/remarkLintTitleCapitalization.js
ilyatitovich: I think the remark has this too. I'll look into the documentation

ai:
I am closing this since we have separated PRs

PR #150, opened 2024-04-05
Add basic pagination for slow feeds

### LINKED ISSUE #136: Add pagination to Slow
Right now having `slow` feeds with 100+ posts freeze browser when you open them.

So we need to add pagination. Requirements:
1. As with pagination in `fast` we need to keep the current page in URL.
2. Adding new posts should not change posts in pages. So it is better to use `since` rather than `page` in URL.
3. In contrast with pagination in `fast` user need to see how many pages are in the list.
4. Users need not only `next page`, but also the last page link, etc.
5. The small `1, 2, 3` links are not very useful. We need something better in terms of UX.

pekac:
Fixes #136 

Added pagination functionality for slow feeds.

Changes made in:
- core/
    - router
    - slow
- web/
    - pages
    - stores

## Checklist

- [x] Run `pnpm test`.
- [x] Think about testing
- If you changed web client:
- If you changed `core/`:

### DIFF
--- a/core/slow.ts
+++ b/core/slow.ts
@@ -9,10 +9,10 @@ import {
 } from './category.js'
 import { client } from './client.js'
 import { onEnvironment } from './environment.js'
-import { BROKEN_FEED, type FeedValue, loadFeed } from './feed.js'
-import { readonlyExport } from './lib/stores.js'
+import { BROKEN_FEED, type FeedValue, loadFeed, loadFeeds } from './feed.js'
+import { listenMany, readonlyExport } from './lib/stores.js'
 import { getPost, getPosts, loadPosts, type PostValue } from './post.js'
-import { router } from './router.js'
+import { type Route, router } from './router.js'
 
 export type SlowCategoriesTree = [CategoryValue, [FeedValue, number][]][]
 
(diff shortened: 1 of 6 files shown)


core/slow.ts
```diff
@@ -120,67 +120,141 @@ let $posts = atom<SlowPostsValue>({ isLoading: true })
...
+
+export type SlowFeedsValue =
+  | { feeds: FeedValue[]; isLoading: false }
+  | { isLoading: true }
+
+let $slowFeeds = atom<SlowFeedsValue>({ isLoading: true })
+
+export const slowFeeds = readonlyExport($slowFeeds)
+
+let POSTS_PER_PAGE = 2
```
ai, line 139: It is very small number :)

ai:
Looks good. Why it is have `Draft` mark?

pekac:
Essentially, because of tests.

Need to check why 1 fails since that part of the logic wasn't touched.
And to ask about UI tests - since this PR adds new state to the UI should I cover that with tests?

ai:
> since this PR adds new state to the UI should I cover that with tests?

Let’§ add visual tests when we will have a good design :)

PR #154, opened 2024-04-07
Improvements for preview server configuration

### LINKED ISSUE #137: Return HTML on apps routes
Right now [dev.slowreader.app](https://dev.slowreader.app) returns app’s HTML only on `/` route. As the result, after opening some page and refreshing the page you will get 404.

In another hand, I do not want to always return app’s HTML. Server should return proper HTTP status 404 on unknown page request.

1. Save web client [routes patterns](https://github.com/hplush/slowreader/blob/0f888be890edbb2ccd630a30607aec67f2fe08b9/web/stores/router.ts#L10-L28) to a variable (don’t forget to add `as const` to keep Nano Stores Router types working) and export it.
2. Add script in `web/scripts/` to generate nginx config to render `/index.html` on request according to those patterns.
3. Call that script on `pnpm build` in `web/`.
4. Use that nginx config in `web/nginx.conf`.
5. Add script description.
6. Add note about that to deploy’s docs.

easing:
## Changes

* Return HTTP 404 if requested path is not in apps routes. Fixes #137
* Add 404.html and 500.html for custom Nginx error pages
* Enable public assets cache for 1 year
* Enable gzip compression (level 1) (initial transfer size reduced from 251Kb to 82Kb)

---

Generating Nginx locations from router definition can be flaky for complex routes, so I've slightly adapt you suggestion of how it can be done and decide to export router RegExps and use [NJS](https://nginx.org/en/docs/njs/) for requests validation.

You wrote in #137:

> In another hand, I do not want to always return app’s HTML. Server should return proper HTTP status 404 on unknown page request.

According to web design basics, it's a [good practice](https://www.nngroup.com/articles/improving-dreaded-404-error-message/) to have non-default 404 page. 

Nginx now returns `index.html` with 404 status if route is not supported by app and `not-found.svelte` template will be rendered. We can use static `404.html` instead, but from my point of view handling not founds in app is easier for developers and gives smoother user experience.

## Small note

At debugging I discovered that `nanostores/router` creates too wide RegExps for patterns like `/feeds/add/:url?` causing path `/feeds/add-random-bug` to be treated as 200 instead of 404, but it's mimics application behaviour (feed adding form will be shown at this path) and seems to be 'technically' correct.

## Checklist

- [x] Run `pnpm test`.
- [x] Think about testing
- If you changed web client:
- If you changed `core/`:


web/.nginx/route-guard.js
```diff
@@ -0,0 +1,32 @@
+// Code in this file is in NJS language https://nginx.org/en/docs/njs/
+
+// eslint-disable-next-line
+let rawRegExps = require('fs').readFileSync(
```
ai, line 4: Why not `import` if we use `export`?

web/README.md
```diff
@@ -33,6 +33,7 @@ We use **Svelte** as the UI framework and **Vite** as the builder.
 - [`.storybook/`](./.storybook/): Storybook’s config.
 - [`index.html`](./index.html): builder entry point. It also contains styles for the app loading state.
 - [`Dockerfile`](./Dockerfile) and [`nginx.conf`](./nginx.conf): web server to serve web client for staging and pull request preview servers.
+- [`.nginx/`](./.nginx/): Nginx related scripts.
```
ai, line 36:
```suggestion
- [`.nginx/`](./.nginx/): preview server related scripts.
```

A good description explain things which is not part of folder name.

easing:
> Wow, it is a bug. Can you send PR to `nanostores/router` if it is easy to fix?

Surely, I'll try to fix it.

> Do we call JS on `/assets/*` requests?

Nginx will try to serve static first and only when no file at requested path was found it will pass request to `@js` location. So JS will called only for missed assets. But we can define `location /assets/  {...}` and avoid JS check for `/assets` at all. Also caching for assets can be added this way. Should we do it in this PR?

```
location /assets {
  expires 1y;
  add_header Cache-Control public;
}
```

---

> But what was a rational reason of not having simple `:param`, `:param?` convert? What is the complexity of that way, can you explain with example?

`:param` substitution should work with current route set, but `nanostores/router` allow us to define route with RegExp and callback (without path pattern). So I prefer not to go this way, because it limits us on how routes can be defined in application.

Also, when creating location RegExp with `:param` convertation, we should handle cases like `|` in path pattern and duplicate part of router logic, so our locations will be out of sync with 'real' router regexes if something will change in `nanostores/router`.

I thinked about dumping regexes from router to Nginx config locations, but for `/^\/path\/(.+)$/.source` JS returns `^\\/path\\/(.+)$` string which will not work in Nginx as we want. Removing `\\` or replacing it with single `\` is [not clear in JS](https://stackoverflow.com/questions/2479309/javascript-and-backslashes-replace) and smells a little bad. And even it can do the job with current routes, there are no garantees that code will work in future when new patterns will be added.

So I choose a more strainforward way to just run checks in JavaScript against the real router rules.

PS: Not important in this case, but Nginx locations order matters and I argue to avoid Nginx locations generation if it's possible.

web/.nginx/route-guard.js
easing: NJS can't import JSON this way. But with fresh mind I realized that we can create module instead of json and switched to module in abdc62c9740c78dd05082d7572c5f2eba1a508bb

ai:
> Should we do it in this PR?

Nice idea. But also add `ui/assets`

ai:
> but nanostores/router allow us to define route with RegExp and callback (without path pattern)

We are not planning to use this feature, so it is OK

> So I choose a more strainforward way to just run checks in JavaScript against the real router rules.

Sure, let’s keep JS way.

ai:
Hm, preview server doesn’t work for some reason

https://preview-154---staging-3ryqvfpd5q-ew.a.run.app/welcome

easing:
> Hm, preview server doesn’t work for some reason

Looks like wrong (default or maybe old) version of Nginx config is used for preview: our custom `404.html` is on server but standard error pages rendered instead of it. https://preview-154---staging-3ryqvfpd5q-ew.a.run.app/404.html

Can you check `/etc/nginx/nginx.conf` in running container? Does it include recent changes? 

Or maybe there is a way for me to pull this preview-154 docker image for investigation?

ai:
> Looks like wrong (default or maybe old) version of Nginx config is used for preview

Yes, it is my mistake. We deploy from `main` branch by sending artifacts of `web/dist/`, but not Dockerfile and routes.

I will fix it separately (need to think about security).

In this case I am waiting for cache changes, and we are ready to merge.

Really smart way to deal with the case.

easing:
> Really smart way to deal with the case.

Glad to hear such comment to my first commit to open source. Thanks.

I've add caching for `/assets` and `/ui/assets` and also enable gzip with minimal compression level. Initial transfer reduced from 251Kb to 82Kb.

With max compression we can save additional 12Kb but CPU will be significantly loaded because responses are compressed on the fly. For max performance without CPU load we should create `.gz` versions of assets on build phase but maybe 12Kb doesn't worth it.

easing:
I've update PR title and description to reflect all changes that has been done

ai:
Thanks. Amazing work.

PR #155, opened 2024-04-07
fix: scroll fast posts after data applying

### LINKED ISSUE #98: Scroll page above on Next page click

sanchopanda:
Fixes #98 

Use tick to wait for the data to be applied before scrolling

## Motivation

Scroll after "mark read and load next" didn't working in firefox

## Checklist

- [ ] Run `pnpm test`.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:

### DIFF
--- a/web/pages/fast.svelte
+++ b/web/pages/fast.svelte
@@ -8,7 +8,7 @@
     openedFastPost,
     fastMessages as t
   } from '@slowreader/core'
-  import { onMount } from 'svelte'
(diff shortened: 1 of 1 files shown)


ai:
Thanks! It was quick!

PR #156, opened 2024-04-08
Add custom remark plugin for checking headings capitalization

### LINKED ISSUE #140: Add more Markdown plugins

ilyatitovich:
Partial #140 

Installed our custom plugin for checking headings.

The plugin works and finds mistakes:
titles

## Motivation

The plugin will allow you to detect mistakes in the capitalization of words in headings.

## Checklist

- [x] Run `pnpm test`.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:

### DIFF
--- a/.remarkrc.js
+++ b/.remarkrc.js
@@ -1,5 +1,9 @@
 // Configuration file used to find popular mistakes in .md files.
 
 export default {
-  plugins: ['remark-lint-no-dead-urls', 'remark-validate-links']
+  plugins: [
+    'remark-lint-no-dead-urls',
+    'remark-validate-links',
+    'remark-lint-heading-capitalization'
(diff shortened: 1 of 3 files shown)


ai:
I fixed `Fixed` to `Partal` in the task description to not close the whole issue on accepting the PR (`Fixes` is a special word).

PR #159, opened 2024-04-10
feat: add search for feed in <a> elements

### LINKED ISSUE #141: Search for feeds in HTML
Some web pages don’t use proper `<link>` tag to their RSS channel.

Instead, they just have `<a>` link.

We need to add to RSS/Atom loader in `core/loaders` small heuristics to search some of the links.

We can use:
1. Link has `feed.` in domain
2. Link has filename `feed.xml`
3. Link has `.rss`, `.atom` extensions
4. Link has `/rss` or `/atom` path

Of course, we will need tests for that.

sanchopanda:
Fixes #141 

added a search for links that meet the following conditions:

1) Link has feed. in domain
2) Link has filename feed.xml
3) Link has .rss, .atom extensions
4) Link has /rss or /atom path

## Motivation

Some web pages don’t use proper <link> tag to their RSS channel.

Instead, they just have <a> link.

## Checklist

- [ ] Run `pnpm test`.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:


ai:
Looks good. But these days I am on the conference. I will try to look at Friday or Saturday.

core/loader/utils.ts
```diff
@@ -25,13 +28,36 @@ function buildFullURL(link: HTMLLinkElement, baseUrl: string): string {
...
+      .map(link => buildFullURL(link, text.url))
+  )
+
+  links.push(
+    ...[...text.parse().querySelectorAll('a')]
+      .filter(a => {
+        let href = a.getAttribute('href')
+        if (!href) return false
+        return (
+          href.includes('feed.') ||
```
ai, line 49: Both Atom and RSS loader will find the same links which is not cool for performance.

Let’s pass regexp for `href` as argument to `findLinks` (like we do with `type`).

sanchopanda: Do you mean that checks on `feed.` and `feed.xml` is enough? I kept it only on `feed.` because it is valid for both cases, I also kept the check for `.atom` and `/atom` because I didn’t understand whether they should be removed or not)

I also added constants for href patterns for <a>, if this is a ok idea, maybe the type for <link> should also be added to the constants, and in findLinks just pass the key as an argument?

core/loader/utils.ts
```diff
@@ -1,11 +1,19 @@
 import type { TextResponse } from '../download.js'
 import type { PreviewCandidate } from '../preview.js'
 
+export const HREF_PATTERNS = {
```
ai, line 4: Why we should have patterns here? Let’s just put string to `loader/rss.ts` and `loader/atom.ts`

core/loader/rss.ts
```diff
@@ -26,7 +26,7 @@ function parsePosts(text: TextResponse): OriginPost[] {
 
 export const rss: Loader = {
   getMineLinksFromText(text, found) {
-    let links = findLinks(text, 'application/rss+xml')
+    let links = findLinks(text, 'application/rss+xml', HREF_PATTERNS.rss)
```
ai, line 29:
```suggestion
    let links = findLinks(text, 'application/rss+xml', /\.rss|\/rss/)
```

core/loader/atom.ts
```diff
@@ -27,7 +27,7 @@ function parsePosts(text: TextResponse): OriginPost[] {
 
 export const atom: Loader = {
   getMineLinksFromText(text, found) {
-    let links = findLinks(text, 'application/atom+xml')
+    let links = findLinks(text, 'application/atom+xml', HREF_PATTERNS.atom)
```
ai, line 30:
```suggestion
    let links = findLinks(text, 'application/atom+xml', /feed\.|\.atom|\/atom/i)
```

core/loader/utils.ts
```diff
@@ -1,11 +1,19 @@
...
+  atom: /feed\.|\.atom|\/atom/i,
+  rss: /feed\.|\.rss|\/rss/i
```
ai, line 6: You don’t need to keep `feed.` in both RegExps. Let’s only Atom look for `feed.`

core/loader/utils.ts
sanchopanda: I suggested this because I use the string in two places, [here](https://github.com/hplush/slowreader/pull/159/files#diff-f3c4dfecd2578ed8df8411f53596fa80abd2ffea573bf8717bd2a5a7f9abcacdR70) and [here](https://github.com/hplush/slowreader/pull/159/files#diff-b25629b68bfa446d7ccfa913ef99c6abfa67d78b8f8823f1715b353937f9c969R30) , but ok, I'll redo it

ai:
Thanks! Great work.

You are right about double places where we use RegExp. Good catch. I think my origin design was wrong and we have duplication there. Wait a little, I will try to fix it.

ai:
Done. Thanks for pointing that out, it helped clean up code a lot.

https://github.com/hplush/slowreader/commit/eafd288c6c852752478da14d57ec05d298b58632

What I did:

1. I separated `getMineLinksFromText` and `getSuggestedLinksFromText`. Now `preview` check that we have no other candidates and call `getSuggestedLinksFromText`, not the loaders itself.
2. I also separate `findLinks` to `findLinksByType` and `findAnchorHrefs`. Previous `findLinks` was useful only in RSS/Atom and will not be useful for upcoming new types of loaders.

PR #160, opened 2024-04-12
[feat] json feed loader

upteran:
Feature #8

- Add json feed loader and tests
- Add to `createTextResponse[download.ts]`  json parse handler
- Add tests with json-feed loader to `preview.test`

## Motivation
- Inside the json feed loader own implementation of the `Loader` methods. Reusing something from the rss/atom loaders probably won't work because the jsonfeed has its own structure and content type
- Update the `createTextResponse` and add a method to handle JSON text. The implementation within the `parse` method will be more complicated because it will add a new text media type. Currently, `parse` returns a document structure, which is why we will need many checks and typings to prevent incorrect type returns.

## Questions

- What order of loaders we should use ? Now it's: atom, jsonfeed, rss 
- 

## Checklist

- [ ] Run `pnpm test`.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:

### DIFF
--- a/core/download.ts
+++ b/core/download.ts
@@ -62,7 +62,10 @@ export function createTextResponse(
     parseJson() {
       let parseType = getContentMediaType(headers)
 
-      if (parseType !== 'application/json') {
+      if (
+        parseType !== 'application/json' &&
+        parseType !== 'application/feed+json'
+      ) {
         return null
       }
 
--- a/core/download.ts
+++ b/core/download.ts
@@ -1,9 +1,11 @@
+import type { JsonFeed } from './loader/json.js'
 import { request } from './request.js'
 
 export interface TextResponse {
   readonly headers: Headers
   readonly ok: boolean
   parse(): Document | XMLDocument
+  parseJson(): JsonFeed | null
   readonly status: number
   readonly text: string
   readonly url: string
@@ -50,6 +52,25 @@ export function createTextResponse(
         }
       }
       return bodyCache
+    },
+    parseJson() {
+      // if (bodyCache) {
+      //   return bodyCache as JsonFeed;
+      // }
+
+      let parseType = headers.get('content-type')
+      if (parseType !== 'application/json') {
+        return null
+      }
+
+      try {
+        let x = JSON.parse(text)
+        return x as JsonFeed
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.error('Parse JSON error', e)
+        return null
+      }
     },
     status,
     text,
--- a/core/download.ts
+++ b/core/download.ts
@@ -1,11 +1,10 @@
-import type { JsonFeed } from './loader/json-feed.js'
 import { request } from './request.js'
 
 export interface TextResponse {
   readonly headers: Headers
   readonly ok: boolean
   parse(): Document | XMLDocument
-  parseJson(): JsonFeed | null
+  parseJson(): null | object
   readonly status: number
   readonly text: string
   readonly url: string
--- a/core/download.ts
+++ b/core/download.ts
@@ -4,7 +4,7 @@ export interface TextResponse {
   readonly headers: Headers
   readonly ok: boolean
   parse(): Document | XMLDocument
-  parseJson(): null | object
+  parseJson(): null | unknown
   readonly status: number
   readonly text: string
   readonly url: string
--- a/core/download.ts
+++ b/core/download.ts
@@ -17,6 +17,14 @@ export interface DownloadTask {
   text(...args: Parameters<typeof request>): Promise<TextResponse>
 }
 
+function getContentMediaType(headers: Headers): string {
+  let contentType = headers.get('content-type') ?? 'text/html'
+  if (contentType.includes(';')) {
+    contentType = contentType.split(';')[0]!
+  }
+  return contentType
+}
+
 export function createTextResponse(
   text: string,
   other: Partial<Omit<TextResponse, 'ok' | 'text'>> = {}
@@ -29,10 +37,8 @@ export function createTextResponse(
     ok: status >= 200 && status < 300,
     parse() {
       if (!bodyCache) {
-        let parseType = headers.get('content-type') ?? 'text/html'
-        if (parseType.includes(';')) {
-          parseType = parseType.split(';')[0]!
-        }
+        let parseType = getContentMediaType(headers)
+
         if (parseType.includes('+xml')) {
           parseType = 'application/xml'
         }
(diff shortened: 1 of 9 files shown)


core/download.ts
```diff
@@ -1,9 +1,11 @@
+import type { JsonFeed } from './loader/json-feed.js'
 import { request } from './request.js'
 
 export interface TextResponse {
   readonly headers: Headers
   readonly ok: boolean
   parse(): Document | XMLDocument
+  parseJson(): JsonFeed | null
```
ai, line 8: `TextResponse` is a general type to load any data.

We should not use exact JSON schema here. Let’s just return the common `object` and type in loader after we checked the format.

core/loader/index.ts
```diff
@@ -16,6 +17,7 @@ export type Loader = {
 
 export const loaders = {
   atom,
+  jsonFeed,
```
ai, line 20: The order is OK

core/loader/json-feed.ts
```diff
@@ -0,0 +1,114 @@
...
+export const jsonFeed: Loader = {
+  getMineLinksFromText(text, found) {
+    let links = [
+      findLinks(text, 'application/feed+json'),
+      // check application/json media type because some resources uses this type instead standard feed+json
+      findLinks(text, 'application/json')
+    ].filter(i => i.length > 0)
+    if (links.length > 0) {
+      // if we have both types of links, we should use feed+json
+      return links[0]!
```
ai, line 84: Too tricky (this is why you need a comment).

It is better to do this avoiding the array.

```js
let links = findLinks(text, 'application/feed+json')
if (links.length === 0) {
  links = findLinks(text, 'application/json')
}
```

ai:
Sorry, we did small refactoring of loaders 😅

See:
https://github.com/hplush/slowreader/pull/159
https://github.com/hplush/slowreader/commit/eafd288c6c852752478da14d57ec05d298b58632

Can you rebase?

core/loader/json-feed.ts
```diff
@@ -0,0 +1,112 @@
...
+  next_url?: string
+  title?: string
+  user_comment?: string
+  version: string
+}
+
+let existJsonFeedVersions = ['1', '1.1']
+
+function isValidJsonFeed(json: unknown): json is JsonFeed {
+  if (typeof json === 'object' && json !== null && 'version' in json) {
```
ai, line 51: It is too simple check for the JSON.

There are many JSONs with `version: '1'` key. Let’s check for some unique items.

core/download.ts
```diff
@@ -4,6 +4,7 @@ export interface TextResponse {
   readonly headers: Headers
   readonly ok: boolean
   parse(): Document | XMLDocument
+  parseJson(): null | object
```
ai, line 7: I just remember, that JSON can be also an array. What do you think if we will use `unknown` here?

core/loader/json-feed.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+  return false
+}
+
+function parsePosts(text: TextResponse): OriginPost[] {
+  let parsedJsonFeed = text.parseJson()
+  if (!isValidJsonFeed(parsedJsonFeed)) return []
+
+  return (
+    parsedJsonFeed.items?.map(item => ({
```
ai, line 63: The app will fall if `items` will be not an array ob objects.

We don’t have this problem in Atom/RSS loader because of XML API.

But since JSONs files in the global Internet could be in any form, we should be ready for broken JSONs.

core/download.ts
upteran: Definitely better decision, you are right, json can be almost any type in current app data

core/loader/json-feed.ts
```diff
@@ -0,0 +1,137 @@
...
+        `Value ${objRecord[field]} of object field ${field} is not valid`
+      )
+      return false
+    }
+  }
+
+  return true
+}
+
+let existJsonFeedVersions = ['1', '1.1']
```
ai, line 75:
```suggestion
const JSON_FEED_VERSIONS = ['1', '1.1']
```

core/loader/json-feed.ts
```diff
@@ -0,0 +1,137 @@
...
+
+let jsonFeedValidationRules: ValidationRules = {
+  items: (value: unknown): boolean => Array.isArray(value),
```
ai, line 78: You don’t need to define types if you defined them in the variable.

Try:

```jsitems: (value) => Array.isArray(value)
```

core/loader/json-feed.ts
```diff
@@ -0,0 +1,137 @@
...
+  title: string
+  user_comment?: string
+  version: string
+}
+
+type ValidationRules = {
+  [key: string]: (value: unknown) => boolean
+}
+
+function isObjValid<ValidatedType>(
```
ai, line 52: What do you think if we rename it is `validate()` to highlight that main step here is rules based validation?

core/loader/json-feed.ts
```diff
@@ -0,0 +1,137 @@
...
+  obj: unknown,
+  rules: ValidationRules
+): obj is ValidatedType {
+  if (typeof obj !== 'object' || obj === null || Array.isArray(obj)) {
```
ai, line 56: You can avoid `as` by using:

```ts
function isObject(value: unknown): value is Record<string, unknown> {
  return typeof obj === 'object' || obj !== null || !Array.isArray(obj)
}
```

core/loader/json-feed.ts
```diff
@@ -0,0 +1,137 @@
...
+    return false
+  }
+
+  let objRecord = obj as Record<string, unknown>
+
+  for (let field in rules) {
+    if (!(field in objRecord) || !rules[field]!(objRecord[field])) {
+      // eslint-disable-next-line no-console
+      console.error(
+        `Value ${objRecord[field]} of object field ${field} is not valid`
```
ai, line 66: If `objRecord[field]` is not a simple string it will be printed as `[Object]`

We can simplify debugging a little by using:

```js
console.error(
    `JSON feed field ${field} is not valid`, objRecord[field]
}
```

Also let’s clarify that it is JSON feed validation.

ai:
Looks very good. If you don’t mind fixing code styling suggestion first, we can merge after that.

ai:
If you have a Twitter name, can you join us Discord and write me there? https://discord.gg/ZUV7JmN9

I want to write about JSON feed support.

core/loader/json-feed.ts
upteran: Hm, I like it. I thought about adding a pointer in name to specify exactly what we validate, but the type and context around probably cover everything, and it's not necessary. I've updated it to just 'validate'.

ai:
A few ideas which came after the merge, which may be interesting for you https://github.com/hplush/slowreader/commit/7f33ce05c5accf181f7a79366a15cdf2e120e8d2

1. `interface` is a little faster than `type`
2. `satisafies` a little better in these contexts, since it is arguments/keys that new types was not added
3. I removed `export type Item` because `Item` is too generic name and could lead to name conflict.
4. `(value): boolean =>` can be replaced to just `(value) =>` because `ValidationRules` already tells what is return value
5. To render non-string value in `console`, you need to pass them as another argument

ai:
Amazing works, thanks!

upteran:
> If you have a Twitter name, can you join us Discord and write me there? https://discord.gg/ZUV7JmN9
> 
> I want to write about JSON feed support.

yep, @pezeze3 in twitter 

thank you for feedback and ideas after, I'll check !

PR #162, opened 2024-04-13
Add favicon for development

### LINKED ISSUE #93: Add icon for development
We should change app favicon for `localhost` and `dev.slowreader.app` to separate dev environment with production.

1. For dev environments we need to create separated versions of SVG icons. We don’t need to duplicate ICO icon.
2. We need to keep light/dark mode `<style>` hack in new icons.
3. Fire on the icons should be with the same yellow color.
4. For development let’s use [campfire](https://pictogrammers.com/library/mdi/icon/campfire/) icon.

Edward-Leopold:
Fixes #93 

- add new icon campfire.svg to /web/main/ folder
- remove link tag with .ico favicon from index.html
- add plugin to vite.config.ts file

# Motivation
I've added transformIndexHtml plugin to vite.config.ts file to track what environment mode the project is in, in order to change the favicon, depending on production or development mode

# Screenshot or Video
image
dev icon with light theme
image
dev icon with dark theme

image
production index.html after building refers to such icon
image
which is for prod dark theme
(similar for the light theme production)

# Checklist

- [ ] Run `pnpm test`.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:


web/index.html
```diff
@@ -11,59 +10,76 @@
     <style>
       :root {
         color-scheme: light dark;
+
```
ai, line 13: Seems like your text editor change the code style.

We need to avoid it. Every PR should focus on single task.

Try to change text editor config and revert changes.

web/index.html
```diff
@@ -2,7 +2,6 @@
 <html>
   <head>
     <meta charset="UTF-8" />
-    <link rel="icon" href="/favicon.ico" sizes="32x32" />
```
ai, line 5: Why you removed it? It will break icon for browsers which has no SVG support.

web/vite.config.ts
```diff
@@ -10,6 +10,18 @@ export default defineConfig(() => ({
     svelte(),
     nodeResolve({
       extensions: ['.js', '.ts']
-    })
+    }),
+    {
+      name: 'html-transform',
+      enforce: 'pre' as const,
```
ai, line 16: To fix ESLint warning call:

```
pnpm eslint --fix .
```

Or change the code according to the error text.

web/vite.config.ts
```diff
@@ -10,6 +10,18 @@ export default defineConfig(() => ({
...
+      transformIndexHtml: (html: string) => {
+        return html.replace(
+          '/main/icon.svg',
+          process.env.NODE_ENV === 'development'
```
ai, line 20: It will use dev icon only in development but now in staging.

We can fix it by define `STAGING` env variable and add it to every `pnpm build` call in `.github/workflows`.

Something like:

```
          process.env.NODE_ENV === 'development' || process.env.STAGING
```

```
  run: pnpm build
  env:
    STAGING: 1
```

It will set `process.env.STANGING`.

You can try it locally by calling `STAGING=1 pnpm build` and then checking the result in `dist/index.html`

web/vite.config.ts
```diff
@@ -10,6 +10,18 @@ export default defineConfig(() => ({
     svelte(),
     nodeResolve({
       extensions: ['.js', '.ts']
-    })
+    }),
+    {
+      name: 'html-transform',
+      enforce: 'pre' as const,
+      transformIndexHtml: (html: string) => {
```
ai, line 17: I think we should mention this is documentation. For instance, Deploy section.

web/vite.config.ts
Edward-Leopold: check .yaml files in .github/workflows, is it okay?

ai: Looks good

web/vite.config.ts
```diff
@@ -10,6 +10,20 @@ export default defineConfig(() => ({
     svelte(),
     nodeResolve({
       extensions: ['.js', '.ts']
-    })
+    }),
+    {
+      enforce: 'pre' as const,
+      // Deploy section.
```
ai, line 16:
```suggestion
```

Let’s remove these comments. In this project we prefer to have it only for complex cases.

web/vite.config.ts
```diff
@@ -10,6 +10,20 @@ export default defineConfig(() => ({
...
+      name: 'html-transform',
+      // Changing the favicon depending on the environment mode
```
ai, line 18: I think the code above is pretty self-describable, we can remove it too.

web/vite.config.ts
```diff
@@ -10,6 +10,20 @@ export default defineConfig(() => ({
...
+      transformIndexHtml: (html: string) => {
```
ai, line 19: We don’t need `: string` type here because `defineConfig()` already defined types for its argument.

So TS will know the types of `transformIndexHtml`.

```suggestion
      transformIndexHtml(html) {
```

And by removing `: string` you can now remove `as const` too. The issue with types was here because we re-define `transformIndexHtml` types.

web/vite.config.ts
```diff
@@ -10,6 +10,20 @@ export default defineConfig(() => ({
...
+        return html.replace(
+          '/main/icon.svg',
```
ai, line 21: It is not cool that we replace SVG icon but keep old ICO.

Also replace `A` to `A` looks like wasting resources.

I suggest:

```js
if (process.env.NODE_ENV === 'development' || process.env.STAGING) {
  return html
    .replace('<link rel="icon" href="/favicon.ico" sizes="32x32" />', '')
    .replace()
} else {
  return html
}
```

In this case for production we will not do anything. And for development we will also remove ICO file.

ai: I found that it works for development, but doesn’t work in staging (this is why you do not see icon in PR preview deployment).

1. In staging Vite replaces `/main/icon.svg` to something like `__VITE_ASSET__Vk_3wFWY__` before running `transformIndexHtml`.
2. Replacing path to `/main/campfire.svg` will not add `campfire.svg` to `/dist` (it works in development because in development Vite doesn’t create `dist/` and look to file-syste, on every HTTP request from browser).

What do I suggest:

1. Replace RegExp to:
   ```js
       .replace(
          /<link rel="icon" href="[^"]+" type="image\/svg\+xml" \/>/,
          '<link rel="icon" href="/campfire.svg" type="image/svg+xml" />'
        )
   ```
   
   This RegExp is ready for `__VITE_ASSET__Vk_3wFWY__` URL.
2. Move `compfire.svg` from `web/main` to `web/public`. In this case Vite will always put it to `dist/`.

ai: How you can check that staging build is working (do it really).

1. Run `STAGING=1 pnpm production`
2. Open browser and check favicon.

PR #163, opened 2024-04-14
Avoid using JS in nginx

ai:
# Motivation

The new way is simpler and should be faster.

Script now exports a single RegExp, which we insert to `nginx.conf` in `Dockerfile`.

cc @easing


easing:
Nice way. I'm surprised that `RegExp.source` works in Nginx without any modification :)

However regexp flags are not exported and final rules will be case-sensitive. To make them case-insensitive (as in `nanostores/router`) Nginx location can be definded this way:

```nginx
location ~* __ROUTES__ {
 #...
}
```

> Regular expressions are specified with the preceding “\~*” modifier (for case-insensitive matching), or the “~” modifier (for case-sensitive matching)
> https://nginx.org/en/docs/http/ngx_http_core_module.html#location

ai:
@easing thanks! Fixed https://github.com/hplush/slowreader/pull/163/commits/5866d13444a4cc05756b1daba8bb516fd4625cd2

PR #164, opened 2024-04-14
Add slow pagination slider-bar

### LINKED ISSUE #152: Add better UI for slow posts pagination
On slow pages we need:

1. An ability to see how many unread posts do we have in total (in this feed).
2. An ability to quickly go to the end and start reading from the end to beginning.
3. An ability to quickly navigate between pages, for instance, to find some posts which should be somewhere in the between.

After discussions in https://github.com/hplush/slowreader/issues/136 we are thinking about progressbar UI like:

```
[----+++---------] 400 posts
[Next page]
```

In progressbar we highlight the current page position and also user can click on any part of slider.

The interesting problem of the task, is that current `since` could not be aligned with `total_unread_posts / per_page`.

pekac:
Fixes #152 

Added a new ui component for pagination that uses input type="range".

# Motivation

To enable quick navigation of posts on slow feed page.

# Checklist

- [ ] Run `pnpm test`.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:


pekac:
@ai, there are 2 main reasons this is in draft:
- need to check tests again
- I used `input type="range"` for the pagination slider - is that ok?

web/ui/pagination-bar.svelte
```diff
@@ -0,0 +1,41 @@
+<script lang="ts">
```
ai, line 1: Can you create a story to check how it works?

Right now it is hard to tell how good `type="range"` fit the case (but I like the care of a11y here).

web/pages/slow.svelte
```diff
@@ -41,12 +43,29 @@
...
+              params: {
+                currentPage: e.detail,
+                feed: $slowFeed
+              },
+              route: 'slow'
+            })
+          }}
+        />
+      {/if}
+      {#if $currentSlowPage < $totalSlowPages}
```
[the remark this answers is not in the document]
pekac: The next-page button you mean?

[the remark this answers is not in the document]
ai: Yes. For me, it looks like this is part of the same control, since they both do the same job.

If we add pagination somewhere else, we will add there Next page button too.

web/pages/slow.svelte
```diff
@@ -41,19 +42,22 @@
...
-        >
-          {$t.showNext}
-        </Button>
+      {/if}
+      {#if $totalSlowPages > 1}
+        <PaginationBar
+          currentPage={$currentSlowPage}
+          label={`${$totalSlowPosts} ${$t.posts}`}
+          totalPages={$totalSlowPages}
+          on:click={e => {
```
ai, line 51: If click change URL, should we use `<a>` instead?

core/router.ts
```diff
@@ -150,22 +150,30 @@ onEnvironment(({ baseRouter }) => {
             }
           }
 
-          if (page.params.since) {
-            if (isNumber(page.params.since)) {
-              let since =
-                typeof page.params.since === 'number'
-                  ? page.params.since
-                  : parseInt(page.params.since)
+          if (page.params.currentPage) {
```
ai, line 153: There is a problem with `currentPage` instead of `since`: if user press `Resfresh` the list will start to jump on every new post loaded. It is OK for the first page, but very strange for others.

Right now we extract all posts from storage. What do you think of keeping `since`?

web/.nginx/routes.js
```diff
@@ -0,0 +1,70 @@
+export default [
```
ai, line 1: Can you delete this file?

It was in `gitignore` but we now save it in another place.

web/stories/ui/pagination-bar.stories.svelte
```diff
@@ -0,0 +1,153 @@
...
+  let currentPage = 1
+  let postsPerPage = 50
+  let totalPosts = 628
+
+  $: totalPages = Math.ceil(totalPosts / postsPerPage)
+  $: postsFrom = (currentPage - 1) * postsPerPage + 1
+  $: postsTo = Math.min(currentPage * postsPerPage, totalPosts)
+</script>
+
+<Story name="Base">
```
ai, line 26: We need more cases for Storybook.

We use it not for manually check, but for automatic snapshots tests. So we need at least tests for:
1. We have only 1 page (no pagination)
2. We have 2 pages
3. We have 3 pages
4. We have 3 pages and the current one is 2
5. We have 3 pages and the current one is the last
6. We have 500 pages

We need all these tests only for the light variant

core/router.ts
pekac: Sure, will revert back to `since`.

There were 2 main reasons for the introduction of `currentPage` and `totalPages`:
1. To make the pagination-bar component possible (e.g. you need to have 10 bars if 10 is the total num of pages and you need 3rd bar marked as active if the user is on page 3)
2. To make clicking on the pagination-bar possible. `since` is great for `next-page` usage, but what if the user wants to click on page 5 from page 1?

For 2. we need the ref to `currentPage` to know which part of the `posts` to slice and if we are already slicing based on `currentPage` we don't have a need for `since` so I decided to remove it. 

However, there's another option - where we can keep `since` and navigate based on it but calculate it from `currentPage` and the `page` clicked upon. 
That initially seemed redundant (keeping everything - `since`, `currentPage`, `totalPages`, etc) so I decided to move `since` out completely and just stick with the `currentPage` implementation. 

Unfortunately, I didn't know about the `Refresh` case :/

ai: I overslept with this idea and seems like your `currentPage `approach is the better option right now.

I am planning to refactor the data system a lot. This is why before this refactoring we should not spend too much effort of pagination (since we will need to rewrite it).

So, let’s keep it and think about refresh case later. Can I just ask you to rename `currentPage` to `page` to keep the short-name code style?

pekac: > I overslept with this idea and seems like your `currentPage `approach is the better option right now.
I mean I just didn't see a way how to support "jump to a specific page" functionality with `since` and without `currentPage`, maybe I'm missing something.

> Can I just ask you to rename `currentPage` to `page` to keep the short-name code style?
Great that you brought it up, I wanted to name it `page` initially, 
Wasn't sure how you would like the `page.params.page` so I switched it to `currentPage`.
Will change it :grin:

ai: > I mean I just didn't see a way to support "jump to a specific page" functionality with since and without currentPage, maybe I'm missing something.

I don’t see a way right now too

web/ui/pagination-bar.svelte
```diff
@@ -0,0 +1,107 @@
...
+  import { createEventDispatcher } from 'svelte'
+
+  import Button from './button.svelte'
+  import Paragraph from './paragraph.svelte'
+
+  export let currentPage: number
+  export let label: string
+  export let totalPages: number
+
+  $: pages = Array.from({ length: totalPages }, (_, i) => i + 1)
```
ai, line 12: Smart way to create an array 💖

Issue #148, opened 2024-04-04
Error when trying to change feed's type


pekac:
Error when trying to change feed's type from "slow" to "fast" from the categories page.

Steps to reproduce:
1. Go to categories page
2. Find a feed that is currently marked as "slow" and has unread posts
3. Change feed's type to "fast"
4. Go to "fast posts" page
5. Feed's posts can not be found on "fast posts" page

Result: 
- Feed's posts not found on "fast posts" page
- Feed found on "slow posts" page

Expected result:
- Feed's posts are visible on "fast posts" page
- Feed should not be visible on "slow posts" page

Video for more info:
![changing-feed-type-crop](https://github.com/hplush/slowreader/assets/7286321/8568eade-4a4f-4aa8-9cf6-ddf13cb9cb23)

ai:
It is more UX issue.

Adding a new filter or changing the category will not move previous posts. The main problem is filters.

Maybe we can load all current posts and change their type on every feed type or filter changes.

chafilin:
I want to take task

PR #167, opened 2024-04-17
Show post on subscribe

### LINKED ISSUE #122: Add post during subscribing to the feed
Now, when you subscribe to the feed you will not see the new feed in Slow/Fast section (because there are no posts after subscription to the feed).

Let’s add the latest post from the new feed.

- [ ] Change `addPreviewCandidate()` in `core/preview.ts` by adding post adding code from [`core/refresh`](https://github.com/hplush/slowreader/blob/main/core/refresh.ts#L109-L124).
- [ ] Find a way to re-use post adding code between `core/preview.ts` and `core/refresh`. For instance, by creating some helper in `core/post.ts`.
- [ ] Add tests

chafilin:
Fixes #122

Added last post after subscribe on feed. 

# Motivation

# Checklist

- [X] Don’t rush. Check all changes in PR again.
- [X] Run `pnpm test`.
- [X] Think about changing documentation.
- [X] If you added a new dependency, check our [requirements](../README.md#dependencies).
- [X] Think about testing
- If you changed web client:
  - [X] Run `pnpm size` and check the difference in the JS bundle size. Is it relevant to the changes? Change the limit in `web/.size-limit.json` if necessary.
  - [X] The UI was checked in Chrome and Firefox (and Safari or Epiphany if you have them).
  - [X] Think about keyboard UX. Is it easy to use the new feature with only one hand on a keyboard? Is it easy to understand what keys to press?
  - [X] Think about accessibility. Check a11y recommendations. Think about how screen reader users will use the tool. Is it easy to use on a screen with bad contrast?
  - [X] Think about translations.
  - [X] Think about right-to-left languages. What parts of the screen should be mirrored for Arabic or Hebrew languages?
  - [X] Think about app loading styles inlined in `index.html`.
- If you changed `core/`:
  - [X] Think about making types more precise. Can you better explain data relations by type?
  - [X] Think about conflict resolution. We don’t need some very smart changing merging; just 2 changes of the same item on different clients should not break the database. What if the user changes an item on one machine and removes it on another?
  - [X] Think about log and storage migration.
  - [X] Change translation ID if you change the meaning of the text.


web/.size-limit.json
```diff
@@ -8,6 +8,6 @@
     "name": "All scripts to execute",
     "path": "dist/assets/*.js",
     "brotli": false,
-    "limit": "210 KB"
+    "limit": "210.04 KB"
```
ai, line 11: Increase it with 5 KB step

ai:
Why Draft PR?

chafilin:
Created draft because have two comments: 
1. I am not sure that helper is needed, because addPost is enough for this case. 
2. I see UX issue when we subscribe to feed and then change filter to slow, but last post already have added to fast

ai:
Thanks! Congrats with the first PR!

ai:
Also, maybe you want to take similar task? https://github.com/hplush/slowreader/issues/148

There we need to change old post’s reading type based on new default feed type or filter changes (re-run filter logic for old posts).

chafilin:
> Thanks! Congrats with the first PR!

Thanks!

> Also, maybe you want to take similar task? #148
> 
> There we need to change old post’s reading type based on new default feed type or filter changes (re-run filter logic for old posts).

Sure, I take it

PR #170, opened 2024-04-17
Fix: Preserve Web page URL or social account handle in input field after reload feed page

### LINKED ISSUE #130: Value data from feeds/add/{data} doesn't passed directly to input where it was previously added
Steps to reproduce:
 - Open http://localhost:5173/feeds/add/
 - Add value to `Web page URL or social account handle` input field (example: https://soundcloud.com/redaxes/flawless-club-mix-feat-echo)
 - Observe how URL is changed regardless of errors (example: http://localhost:5173/feeds/add/https%3A%2F%2Fsoundcloud.com%2Fredaxes%2Fflawless-club-mix-feat-echo)
 - Refresh the page - value from URL is not returned to input box

kmpeeduwee:
Fixes #130 

# Motivation

# Screenshot or Video

# Checklist

- [ ] Run `pnpm test`.
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:

### DIFF
--- a/core/preview.ts
+++ b/core/preview.ts
@@ -75,7 +75,15 @@ export const previewCandidates = computed($candidates, candidates => {
 })
 
 onMount($candidates, () => {
-  return $links.listen(() => {})
+  let unsubscribeRouter = router.subscribe(page => {
+    if (page.route === 'add') {
+      setPreviewUrl(page.params.url!)
+    }
+  })
+
+  return () => {
+    unsubscribeRouter()
+  }
 })
 
 let $candidate = atom<string | undefined>()


core/preview.ts
```diff
@@ -75,7 +75,15 @@ export const previewCandidates = computed($candidates, candidates => {
 })
 
 onMount($candidates, () => {
-  return $links.listen(() => {})
+  let unsubscribeRouter = router.subscribe(page => {
```
ai, line 78: You can just:

```js
return router.subscribe(page => {
```

core/preview.ts
```diff
@@ -75,7 +75,15 @@ export const previewCandidates = computed($candidates, candidates => {
...
+    if (page.route === 'add') {
+      setPreviewUrl(page.params.url!)
```
ai, line 80: We already have `setPreviewUrl()` call on `router` changes https://github.com/hplush/slowreader/blob/main/core/preview.ts#L319

Maybe it will be better to fix that block to avoid code duplication?

ai: Can I ask you to do a deep investigation:

1. Add `console.log` with different variables to that `router.listen`
2. Reproduce an issue
3. Find a reason why exactly the previous code didn’t work
4. Explain it in fix PR

It is very dangerous to fix an issue without going deep into why it happened.

We need something like:
> We have a issue because `previewUrl` was set on page loading event, so I added extra variable to track page loading event

ai:
I need this fix to debug another issue, so I fixed it by my own.

You were close, just didn’t reuse another block where we subscribed to the store.

Here is a better way: 506c628

PR #172, opened 2024-04-21
Update feed type

### LINKED ISSUE #148: Error when trying to change feed's type

chafilin:
Fixes #148 

Change posts reading after change feed reading

# Checklist

- [X] Don’t rush. Check all changes in PR again.
- [X] Run `pnpm test`.
- [X] Think about changing documentation.
- [X] If you added a new dependency, check our [requirements](../README.md#dependencies).
- [X] Think about testing
- If you changed web client:
  - [X] The UI was checked in Chrome and Firefox (and Safari or Epiphany if you have them).
  - [X] Think about keyboard UX. Is it easy to use the new feature with only one hand on a keyboard? Is it easy to understand what keys to press?
  - [X] Think about HTML semantics.
  - [X] Think about accessibility. Check a11y recommendations. Think about how screen reader users will use the tool. Is it easy to use on a screen with bad contrast?
  - [X] Think about translations.
  - [X] Think about right-to-left languages. What parts of the screen should be mirrored for Arabic or Hebrew languages?
  - [X] Think about app loading styles inlined in `index.html`.
- If you changed `core/`:
  - [X] Think about making types more precise. Can you better explain data relations by type?
  - [X] Think about conflict resolution. We don’t need some very smart changing merging; just 2 changes of the same item on different clients should not break the database. What if the user changes an item on one machine and removes it on another?
  - [X] Think about log and storage migration.
  - [X] Change translation ID if you change the meaning of the text.


core/feed.ts
```diff
@@ -100,6 +100,16 @@ export async function changeFeed(
   feedId: string,
   changes: Partial<FeedValue>
 ): Promise<void> {
+  if (changes.reading) {
```
ai, line 103: We need to change post after changing the feed.

Imagine if we will add new post in the middle of this process.

core/post.ts
```diff
@@ -66,6 +67,13 @@ export function deletePost(postId: string): Promise<void> {
   return deleteSyncMapById(getClient(), Post, postId)
 }
 
+export async function changePost(
```
ai, line 70: Can you add a simple test for that? You can just add a few lines in existing test.

core/feed.ts
```diff
@@ -100,6 +100,16 @@ export async function changeFeed(
...
+    let posts = await loadPosts({ feedId })
+    await Promise.all(
+      posts.map(post =>
+        changePost(post.id, {
+          reading: changes.reading
```
ai, line 108: Post’s type is not only the result of feed type, but also a result of filters.

This code will override filters. We need to use them here and also re-run the code on filter’s changes too. Maybe it will be a good idea to move post’s type calc to some function.

https://github.com/hplush/slowreader/blob/main/core/refresh.ts#L109-L117

core/test/feed.test.ts
```diff
@@ -111,3 +111,18 @@ test('shows that user has any feeds', async () => {
...
+  let feedId = await addFeed(testFeed())
+  let feed = getFeed(feedId)
+  let posts = getPosts()
+  keepMount(feed)
+  keepMount(posts)
+  await addPost(testPost({ feedId }))
+
+  equal(ensureLoaded(feed.get()).reading, 'fast')
+  equal(ensureLoaded(posts.get()).list[0]?.reading, 'fast')
+  await changeFeed(feedId, { reading: 'slow' })
```
ai, line 125: Can you add filter to the test?

Let’s we have 2 old posts (one with type by filter, other with default type).

core/post.ts
```diff
@@ -94,6 +104,73 @@ export function getPostIntro(post: OriginPost): string {
   }
 }
 
+export async function calcPostReading(
+  post: OriginPost,
+  feed?: FeedValue,
+  filterId?: string
+): Promise<'fast' | 'slow'> {
+  let checker = prepareFilters(
+    await loadFilters({
```
ai, line 113: Loading filters again for every posts in feed can be very slow.

For instance, in `changePostsByFeed` you can prepare `checker` once and use it for all posts.

core/post.ts
```diff
@@ -94,6 +104,73 @@ export function getPostIntro(post: OriginPost): string {
...
+export async function changePostsByFeed(feedId: string): Promise<void> {
+  let posts = await loadPosts({ feedId })
+  let feed = await loadValue(await getFeed(feedId))
+  for (let post of posts) {
+    let reading = await calcPostReading(post, feed)
+    await changePost(post.id, { reading })
+  }
+}
+
+export async function changePostsByFilter(filterId: string): Promise<void> {
```
ai, line 150: Do we really need separated function for `changePostsByFeed` and `changePostsByFilter`?

In all cases we need to:
1. load all filters (because filters order is important)
2. load all posts
3. re-calc post type according to filters and default type

I suggest create functions like `recalcPostsReading(feedId)`.

core/filter.ts
```diff
@@ -92,14 +94,25 @@ export async function addFilterForFeed(feed: FeedValue): Promise<string> {
 }
 
 export async function deleteFilter(filterId: string): Promise<void> {
-  return deleteSyncMapById(getClient(), Filter, filterId)
+  let filter = await loadValue(Filter(filterId, getClient()))
+  let feedId = filter?.feedId
```
ai, line 98: Let’s just wrap this code in

```js
if (filter) {
}
```

It will be cleaner.

core/filter.ts
```diff
@@ -92,14 +94,25 @@ export async function addFilterForFeed(feed: FeedValue): Promise<string> {
...
+  }
 }
 
 export async function changeFilter(
   filterId: string,
   changes: Partial<FilterValue>
 ): Promise<void> {
-  return changeSyncMapById(getClient(), Filter, filterId, changes)
+  await changeSyncMapById(getClient(), Filter, filterId, changes)
+  let filter = await loadValue(Filter(filterId, getClient()))
```
ai, line 111: The same here, just wrap it.

core/filter.ts
chafilin: Is it save to wrap `deleteSyncMapById` in `if(filter)`?
And can the filter object be deleted during `deleteSyncMapById`? 
In that cases it can be dangerous to wrap it

ai:
Perfect. It will improve UX a lot!

PR #174, opened 2024-04-23
Loaders smoke tests on real content

### LINKED ISSUE #10: Loaders smoke tests on real content
**Blocked by https://github.com/hplush/slowreader/issues/96**

In `loader-tests/` we need to create a config with URLs to test loaders on real data.

- [ ] Add script file and `integration` script to project’s `package.json`. It will use a lot from #96, think about code re-using.
- [ ] This script need to read `loader-tests/feeds.yml`
- [ ] Add there `dev.to`, LiveJournal, Tumblr (some safe-for-work project, ideal the artist with their work), WordPress official blog, other blog engines blogs, some popular blogs you know (Github).
- [ ] For each RSS try to find the right loader by calling `loader.isMineText()`. Show nice error if the loader was not found.
- [ ] For found loader call `loader.getMineLinksFromText()`. Be sure that it returned some posts.
- [ ] For URL remove path (`https://example.com/news.rss` → `https://example.com/`) and call use methods from `core/preview.ts` to be sure that reader will find that RSS URL from the home. `feeds.yml` should have a marker to disable this test for feed.
- [ ] Add to docs information about that test and that file.
- [ ] Add CI workflow to run that script every day.

ilyatitovich:
Fixes #10 

- Added `feeds.yml` file
- Added script for testing feeds from `feeds.yml`
- For handle same tasks like in #96, the code from `check-opml.ts` was reused (moved to `utils.ts`)
- Added scripts to `package.json`
- Added CI workflow to run daily feeds check
- Added information about the test in docs

# Motivation

The test will allow us to check loaders on various types of popular blogging platforms.

# Checklist

- [x] Run `pnpm test`.
- [x] Think about testing
- If you changed web client:
- If you changed `core/`:


loader-tests/README.md
```diff
@@ -14,3 +14,15 @@ Test that Slow Reader can work with all feeds from your RSS reader by using OPML
    ```sh
    pnpm check-opml PATH_TO_YOUR_FILE.opml
    ```
+
+## Check Loaders by Different Blog Platforms
+
+Test that Slow Reader can work with different feeds from popular blogging platforms.
+
+1. Check out [`feeds.yml`](./feeds.yml) for the structure of `.yml` file.
+2. Add your favorite blogging platforms.
```
ai, line 23: Here we need only to explain how to run tests and what they did.

loader-tests/check-feeds.ts
```diff
@@ -0,0 +1,36 @@
+import yaml from 'yaml'
+
+import {
+  enableTestClient,
+  error,
+  type Feed,
```
ai, line 6: Don’t like that we have an export (not an internal type as before) with name conflict with `core`’s `Feed`.

Can we rename it?

loader-tests/package.json
```diff
@@ -3,12 +3,14 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "check-opml": "tsx check-opml.ts"
+    "check-opml": "tsx check-opml.ts",
+    "check-feeds": "tsx check-feeds.ts feeds.yml"
```
ai, line 7: What do you think of renaming it to `test-loaders` since we are not testing the feeds (websites).

.github/workflows/test-loaders.yml
```diff
@@ -0,0 +1,17 @@
+name: Daily Loaders Test
+on:
```
ai, line 2: Let’s also run on `.github/workflows/test-loaders.yml` or related TS files changes.

Debugging cron-based workflows are hard (because you need to wait a day until they will run). Adding `pull_request` and `push` with `paths` (like [here](https://github.com/hplush/slowreader/blob/main/.github/workflows/docs.yml)) can simplify testing changes.

ai: Maybe we should also run tests in commits/PRs chinging `core/loader/*.ts`? What do you think?

loader-tests/README.md
```diff
@@ -14,3 +14,14 @@ Test that Slow Reader can work with all feeds from your RSS reader by using OPML
    ```sh
    pnpm check-opml PATH_TO_YOUR_FILE.opml
    ```
+
+## Check Loaders by Different Blog Platforms
+
+Test that Slow Reader can work with different feeds from popular blogging platforms.
+
+1. Check out [`feeds.yml`](./feeds.yml) for the structure of `.yml` file.
```
ai, line 22: Just remove the `ol` :)

.github/workflows/test-loaders.yml
ilyatitovich: I agree, additional test for updated loaders on real data is a great idea. If developer will forget to run test locally it is run by the way automatically. I will add paths for push and РR.

ai:
Thanks. Fast and good work.

PR #186, opened 2024-04-29
<TwoStepPage> mobile support

### LINKED ISSUE #95: Add mobile support to `<TwoStepsPage>` and Back button to mobile navbar
`<TwoStepsPage>` is used to show list on the left and selected list item on the right.

For instance, we use it in [feeds page](http://localhost:5173/feeds/categories) with feeds list on the left and selected feed on the right. On in the slow/fast feed with posts list on the left and selected post on the right.

On the phone we can’t show 2 sides, so we need to show only one side (list first and item when we chose it).

- [ ] Add a way to detect what side we need to show. We can use router for that. For instance, we can add `backRoutes[routeName] => (currentRoute) => backRoute` object.
- [ ] Think about some global API in `core` with what side is showed. We will need it in `<Navbar>` next. For instance, `backRoutes[route.get().route]`
- [ ] Add back button to Navbar which will show the list side back. You should replace Refresh icon with Back button when item’s side is shown. 
- [ ] It will be nice to add <kbd>Esc</kbd> support and to switch from item side to list side on pressing <kbd>Esc</kbd>.
- [ ] The same Exit button for desktop to the `<TwoSidePage>` (to the top of right panel).
- [ ] Add stories to test it and unit tests to `core` changes.
- [ ] Try to move to `core` as much as possible since we may need the same logic in React Native client later.

We also need to think about scroll position on list—item switching.

The task is blocked by https://github.com/hplush/slowreader/issues/94

ilyatitovich:
Fixes #95 

**Done:**
- Added logic to core - check which side should be displayed depending on the path. Next -  manipulate CSS classes to show/hide UI elements.
- Added `Back` button to <Navbar> (mobile version).
- Added `Esc` key support for switching between sides
- Added utit tests with one exception:

```js
test('show second side: add/url', async () => {
  setBaseTestRoute({ params: { url: 'dev.to/feed' }, route: 'add' })
  await setTimeout(100)
  strictEqual(side.get(), 'second')
})

```
Test gives an error: `TypeError [Error]: request is not a function`. I haven't figured out why yet.

**In Progress:**
- Add stories for UI testing.
- I didn’t find `<TwoSidePage>`, maybe I understood something wrong 😀
- Now, scroll position when switching between sides is preserved.

# Motivation

These changes will improve the user experience on mobile devices.

# Checklist

- [x] Run `pnpm test`.
- [x] Think about testing
- If you changed web client:
- If you changed `core/`:


ai:
> I didn’t find <TwoSidePage>, maybe I understood something wrong

[`TwoStepsPage`](https://github.com/hplush/slowreader/blob/main/web/ui/two-steps-page.svelte) 😅

> Package size limit has exceeded by 1.34 kB

Increase size limit with 5 KB step

> Test gives an error: TypeError [Error]: request is not a function. I haven't figured out why yet.

You need to mock the request like in https://github.com/hplush/slowreader/blob/main/core/test/preview.test.ts

1. `mockRequest()` in `beforeEach` 
2. `checkAndRemoveRequestMock()` in `afterEach`
3. `expectRequest` with expected URLs in the test

We can add some simpler mock if needed.

> setBaseTestRoute({ params: { url: 'dev.to/feed' }, route: 'add' })
> await setTimeout(100)
> strictEqual(side.get(), 'second')

How user will select feed if we switch to the second step? The feed adding process:

1. Enter URL in first step
2. Select feed candidate in first step
3. Only then preview feed posts and add feed in the second step

Maybe route-based logic is not enough for all two-step logic. We can export some methods from two-steps to use in `preview` and similar pages when steps are separated by internal state, not route.

ilyatitovich:
Thanks! Now process to add feed is clearer. I'll add some function to check this steps.

ilyatitovich:
- Updated logic to display sides when adding a feed.
- Added Story: Navbar with Back button. 
- Added Story: Adding feed process on mobile screen.

core/two-steps.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export type Side = 'first' | 'second'
+
+export const backRoutes: BackRoutes = {
+  categories: '/feeds/categories',
+  fast: '/fast/:category',
+  slow: '/slow/:feed'
+}
+
+export const side = atom<Side>('first')
```
ai, line 16: Maybe we can use boolean since we do not plan third page?

It could simplify code and types.

We can rename store to `secondStep`

core/two-steps.ts
```diff
@@ -0,0 +1,45 @@
+import { router } from '@slowreader/core'
+import { atom } from 'nanostores'
+
+interface BackRoutes {
+  [key: string]: string
+}
+
+export type Side = 'first' | 'second'
+
+export const backRoutes: BackRoutes = {
```
ai, line 10: URL is not part of `core`, we can’t keep URLs patterns here

core/two-steps.ts
```diff
@@ -0,0 +1,45 @@
+import { router } from '@slowreader/core'
+import { atom } from 'nanostores'
+
+interface BackRoutes {
```
ai, line 4: `Record<A, B>` is a great helper to create `interface { [key: A]: B }`

core/two-steps.ts
```diff
@@ -0,0 +1,45 @@
...
+        side.set('first')
+    }
+  }
+}
+
+router.subscribe(route => {
+  if (route.route === 'categories' && route.params.feed) {
+    side.set('second')
+  } else if (route.route === 'fast' && route.params.post) {
+    backRoutes.fast = `/fast/${route.params.category}`
```
ai, line 37: We can’t use URL here too.

It is better to use `envrionment.openRoute()` which accept just a route object (and environment use URL or any other type of routing).

core/two-steps.ts
```diff
@@ -0,0 +1,45 @@
...
+export const backRoutes: BackRoutes = {
+  categories: '/feeds/categories',
+  fast: '/fast/:category',
+  slow: '/slow/:feed'
+}
+
+export const side = atom<Side>('first')
+
+export function toggleSide(event: KeyboardEvent): void {
+  if (event.key === 'Escape') {
```
ai, line 19: I am not sure that this code can be used in desktop app or iOS app, since they will not have `KeybordEvent`.

web/ui/two-steps-page.svelte
```diff
@@ -23,18 +24,26 @@
   }
 
   onMount(() => {
+    window.addEventListener('keydown', toggleSide)
```
ai, line 27: I think we can move `keycode === 'Escape'` code here.

But we always should check `document.active` in any hot key listener. User focus can be in text field or other control, where `Esc` should have different meaning.

web/ui/two-steps-page.svelte
```diff
@@ -23,18 +24,26 @@
...
       document.title = prevTitle
     }
   })
 </script>
 
 <main id="page" class="two-steps-page">
-  <div bind:this={first} class="two-steps-page_step">
+  <div
+    bind:this={first}
+    class={`two-steps-page_step ${$side === 'second' ? 'two-steps-page_step-desktop-only' : ''}`}
```
ai, line 39: In BEM, you can’t put 2 elements on the same tag.

I think `two-steps-page_step-desktop-only` should be a modifier like `is-*`.

And the current name is not clear for me. Is it `is-current`?

web/ui/navbar/index.svelte
```diff
@@ -49,28 +51,45 @@
 
 <nav class="navbar">
   <div class="navbar_main" aria-orientation="horizontal" role="menu">
-    {#if $isRefreshing}
+    <div
+      class={`navbar_back-btn ${$side === 'first' ? 'navbar_btn-hidden' : ''}`}
```
ai, line 55: `is-hidden` modifier should be here instead of 2 elements on the same tag

web/ui/navbar/index.svelte
```diff
@@ -49,28 +51,45 @@
...
-      </NavbarItem>
-    {:else}
-      <NavbarItem
-        name={$t.refresh}
-        current={$router.route === 'refresh'}
-        hotkey="r"
-        icon={mdiRefresh}
+        name={$t.back}
+        current={true}
+        hotkey="Esc"
```
ai, line 60: Will we have 2 back-to-first step calls on small screen on pressing `Esc`?

The first one will be from `<TwoStepPage>` listener, the second will be from the button (`hotkey` is not just a visual, it creates listener).

web/ui/two-steps-page.svelte
ai: See `keyux` source code for details https://github.com/ai/keyux/blob/main/hotkey.js

web/ui/navbar/index.svelte
```diff
@@ -49,28 +51,45 @@
...
+        href={backRoutes[$router.route]}
+        icon={mdiChevronLeft}
         small
         on:click={() => {
-          refreshPosts()
+          !backRoutes[$router.route] && side.set('first')
```
ai, line 65: `if` here will be more readable than `condition && action()`

web/ui/navbar/index.svelte
```diff
@@ -49,28 +51,45 @@
...
-    {:else}
-      <NavbarItem
-        name={$t.refresh}
-        current={$router.route === 'refresh'}
-        hotkey="r"
-        icon={mdiRefresh}
+        name={$t.back}
+        current={true}
+        hotkey="Esc"
+        href={backRoutes[$router.route]}
```
ai, line 61: To get back URL here you can do something like: 

```js
getPagePath(urlRouter, backRoute)
```

backRoute will be from the core (which know nothing about URLs) and here we will combine route with web's URL router which has logic to convert abstract routes to URL.

web/ui/navbar/index.svelte
ilyatitovich: `Esc` is not valid name for key code, so it will no affect on `Back` button:
```js
let press = findHotKey(event, window, transformers)
if (press) press.click()
```

ai: In this case we should not use it, since invalid data will confuse other developers.

it is OK just remove `hotkey` here.

core/two-steps.ts
ilyatitovich: If we will create a client for iOS on React Native, isn’t there `KeyboardEvent`? Will the desktop version be on Electron.js?

ai: I am planning full native clients. For instance, for GNOME, I am planning to use GTK bindings to JS.

This is why we separate logic between core and clients.

ilyatitovich: Should we implements separate `Escape` button support for each platform?

ai: We can listen for Escape button in client code, then listener will run core's function.

web/ui/navbar/index.svelte
ilyatitovich: Could you give me an example of valid CSS class name that I can use here? My 'names' is always give an error when I run test:
```bash
Selector .is-hidden does not follow our BEM name system
.is-hidden {
```

ai: Because every modifier need to be defined with element.

```css
navbar_back-button.is-hidden {}
```

ilyatitovich:
Removed URL patterns from the `core` code and updated logic.

core/test/two-steps.test.ts
```diff
@@ -0,0 +1,27 @@
...
+beforeEach(() => {
+  enableClientTest()
+})
+
+afterEach(async () => {
+  await cleanClientTest()
+})
+
+test('show first side', () => {
+  setBaseTestRoute({ params: {}, route: 'add' })
```
ai, line 16: If `router.subscribe()` doesn’t work, we will have the same `false`.

We could improve test by something like:

```js
  setBaseTestRoute({ params: {}, route: 'add' })
  strictEqual(secondStep.get(), false)

  showSecondStep() 
  strictEqual(secondStep.get(), false)

  setBaseTestRoute({ params: {}, route: 'categories' })
  strictEqual(secondStep.get(), false)
```

web/ui/two-steps-page.svelte
```diff
@@ -22,19 +23,33 @@
     })
   }
 
+  function toggleStepsOnEscape(event: KeyboardEvent): void {
+    if (event.key === 'Escape' && document.activeElement?.tagName === 'BODY') {
+      toggleSteps()
```
ai, line 28: Do you have a reason why on `Esc` on the first step we should move to the second?

web/pages/feeds/add.svelte
```diff
@@ -52,6 +53,7 @@
               current={$previewCandidate === candidate.url}
               on:click={() => {
                 setPreviewCandidate(candidate.url)
+                showSecondStep()
```
ai, line 56: We should put maximum code to the `core`.

Let’s put it to `setPreviewCandidate`.

web/pages/feeds/categories.svelte
```diff
@@ -85,6 +86,7 @@
                         params: { feed: feed.id },
                         route: 'categories'
                       })}
+                      on:click={showSecondStep}
```
ai, line 89: Calling `showSecondStep` on click is not a good idea, because URL is the source of truth, not clicks.

Here is a use case when it will not work:

1. I open categories page
2. I click on feed A
3. Slowreader open feed A and I see Back button in the menu
4. I press Back button *in the browser*
5. I still see Back button in navbar

web/ui/two-steps-page.svelte
ilyatitovich: No, I just thought I should implement switching between sides using the Escape key.

ai: Let’s use `showFirstStep()` here.

I'm afraid that we will have unexpected behavior of switching 1→2 on `Esc` key.

web/pages/feeds/categories.svelte
ilyatitovich: I will try to figure out how to implement back routes.

core/two-steps.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function showSecondStep(): void {
+  secondStep.set(true)
+}
+
+export function showFirstStep(): void {
+  secondStep.set(false)
+}
+
+router.subscribe(route => {
```
ai, line 14: The logic looks similar to `backRoute`. What if we will replace `route.route === 'categories'` to `route.route === 'categories' && route.params.feed` in `backRoute` and here will use:

```js
backRoute.subscribe(url => {
  if (url) {
      showSecondStep()
  } else {
      showFirstStep()
  }
})
```

web/ui/navbar/index.svelte
```diff
@@ -49,28 +52,43 @@
...
-      </NavbarItem>
-    {:else}
-      <NavbarItem
-        name={$t.refresh}
-        current={$router.route === 'refresh'}
-        hotkey="r"
-        icon={mdiRefresh}
+        name={$t.back}
+        current={true}
+        hotkey="Escape"
```
ai, line 59: There 2 problems with `Esc` handling right now:

1. Button is too small for `Escape` hotkey preview
Captura desde 2024-05-13 12-54-35
2. The `Esc` handling works unpredicted on desktop.

Case 1:
1. Open desktop mode
2. Add `github.blog` as Slow
3. Open Slow
4. Open GitHub
5. Click on Open post
6. Press Esc → Post will be closed (Great), focus will be returned to posts list (not so good, since we had no focus, but OK)
7. Press Esc again → focus can’t be disabled anymore and always return to the posts list (expected behaviour: no focus)

Case 2:
1. Open add
2. Enter `github.blog`
3. Press Enter → focus moves to the links list
4. Press Enter → focus moved to the side bar
5. Press Esc → focus return to the field, not to the list as expected

Case 3:

1. Add `github.blog`
2. Open By categories
3. Click on GitHub → side panel will be open
4. Press Esc → side panel will not be closed (until I press Esc again)

I assume there is Esc conflict with `keyux` jumps. Let's remove `hotkey="Escape"` and do right Escape handling as separated task.

ilyatitovich:
- Updated `backRoute` code
- Removed `Esc` hotkey

ai:
Thanks!

PR #187, opened 2024-04-29
Navbar progress semantics

### LINKED ISSUE #86: Improve semantic of navbar progress
We have SVG with post uploading progress in menu.

Captura desde 2024-03-18 22-49-02

Unfortunately, it is not very accessible from screen readers because it has no semantics about the current progress.

1. We need to add semantics [using this guide](https://piccalil.li/blog/solution-009-progress-indicator/).
2. It will be good to double-check in other a11y guides.

AnnaGolmakova:
Fixes #86

Added role=progressbar and aria-label for loading indicator in navigation

# Motivation

To help with accessibility

# Checklist

- [ ] Run `pnpm test`.
- [ ] If you added a new dependency, check our [requirements](https://github.com/hplush/slowreader/blob/main/README.md#dependencies).
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:


core/messages/navbar/en.ts
```diff
@@ -8,6 +8,7 @@ export const navbarMessages = i18n('navbar', {
   feeds: 'Manage Feeds',
   feedsByCategory: 'By Category',
   interface: 'User Interface',
+  loading: 'Loading',
```
ai, line 11: Let’s re-use it from common messages: https://github.com/hplush/slowreader/blob/main/core/messages/common/en.ts#L7

ai, line 11: Let’s explain a little what is loadings. For instance:

```js
postRefreshing: 'Checking feeds for new posts'
```

ai:
Now there is only code style error from ESLint.

To fix it, just call:

```sh
pnpm eslint --fix .
```

It will fix all ESLint warnings (which is possible to fix automatically).

ai:
Thanks! Now our a11y was improved!

Issue #195, opened 2024-05-10
Add non-strict XML parser


ai:
Sometimes feeds XML contains small syntax errors like [here](https://uxcellence.com/rss/):

```
error on line 6 at column 87: Namespace prefix atom on link is not defined
```

We need to ignore similar types of issue.

Could be related to https://github.com/hplush/slowreader/issues/9

ai:
Combine tasks into https://github.com/hplush/slowreader/issues/323

PR #196, opened 2024-05-12
Add progress bar to loader-tests

### LINKED ISSUE #143: Add progress to Check OPML script
We have `loader-tests/check-opml.ts` script to check feeds from OPML file. You can run it by:

```sh
pnpm check-opml ./loader-tests/example.ompl
```

We want to add progress bar to the process like:

```
███████████░░░░░
```

In `loader-tests` we plan to have many scripts, so this progress bar should re-usable from `loader-tests/utils.ts`. For instance:
1. In the start script define jobs number by `setJobs(number)`
2. On every `success()`/`error()` call we mark next job as completed.

We should not show progress on CI.

Kamino0:
Fixes #143
Added progress bar to loader-tests.

Also changed `&` to `&amp;` in example.opml to satisfy parser.
Recently there was a [fix](https://github.com/hplush/slowreader/commit/ef1cea249e5e4d50198167c41fac90a1f426fe99) to the сontentType definition and on my computer [queryselectorAll](https://github.com/hplush/slowreader/blob/main/loader-tests/check-opml.ts#L22) started returning an empty list.
I think the problem in more strict parsing of XML in comparison with HTML, since previously the example.opml file was defined as text/html and now as application/xml.
I suppose in future resolving issue #195 fix this problem.

progress-bar

# Checklist

- [x] Run `pnpm test`.
- [x] Think about testing
- If you changed web client:
- If you changed `core/`:

### DIFF
--- a/loader-tests/check-opml.ts
+++ b/loader-tests/check-opml.ts
@@ -8,10 +8,10 @@ import {
   fetchAndParsePosts,
   findRSSfromHome,
   finish,
+  initializeProgressBar,
   isString,
   type LoaderTestFeed as OpmlFeed,
-  readText,
-  initializeProgressBar
+  readText
 } from './utils.js'
 
 async function parseFeedsFromFile(path: string): Promise<OpmlFeed[]> {
@@ -61,7 +61,7 @@ cli.run(async args => {
 
   let feeds = await parseFeedsFromFile(opmlFile)
 
-  const jobsCount = home ? feeds.length * 2 : feeds.length
+  let jobsCount = home ? feeds.length * 2 : feeds.length
   initializeProgressBar(jobsCount)
 
   await completeTasks(
--- a/loader-tests/check-opml.ts
+++ b/loader-tests/check-opml.ts
@@ -10,7 +10,8 @@ import {
   finish,
   isString,
   type LoaderTestFeed as OpmlFeed,
-  readText
+  readText,
+  initializeProgressBar
 } from './utils.js'
 
 async function parseFeedsFromFile(path: string): Promise<OpmlFeed[]> {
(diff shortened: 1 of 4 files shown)


ai:
Can you show screenshot of progress bar?

loader-tests/check-opml.ts
```diff
@@ -59,6 +60,10 @@ cli.run(async args => {
   }
 
   let feeds = await parseFeedsFromFile(opmlFile)
+
+  const jobsCount = home ? feeds.length * 2 : feeds.length
```
ai, line 64: Run `eslint --fix .` to fix code style.

And you marked `Run pnpm test` as done, but didn’t call it ;).

loader-tests/check-opml.ts
```diff
@@ -10,7 +10,8 @@ import {
   finish,
```
ai, line 10: We need progress bar also in `test-loaders`

Kamino0:
> Can you show screenshot of progress bar?

Added gif to main comment

loader-tests/check-opml.ts
Kamino0: done

loader-tests/check-opml.ts
Kamino0: fixed 👀

Kamino0: I called, but it seems these errors were hidden under a lot of 
```
Unable to read file core/coverage/tmp/coverage-49309-1715592007988-0.json
ENOENT: no such file or directory, open /core/coverage/tmp/coverage-49309-1715592007988-0.json
```

Want look at this errors later

ai:
Thanks! Great job.

PR #202, opened 2024-05-15
Move from asdf to Dev Container

ai:
Right now we use `asdf` to sync environment. It has a few problems:

1. `asdf` has conflict with `nvm`. The conflict behavior is very strange and user can spend a lot of time finding the source of issue.
2. Some tool in `node_modules` may contain malware. It should not have full access to user’s home folder with credentials JSONs and private files. Dev Container isolate project.
3. Every user will have exactly the same environment.

Risks:
1. I test it only for Podman. I need Docker user to test it.
2. We need to test Windows and Mac performance.


baileys-li:
> We need to test Windows and Mac performance.
I have both

@ai What do I need to test? Try `pnpm test`. In the dev container, the performance slows down perceptibly. You can work, but you can feel the difference.

#### macOS M1 in dev container:
pnpm test  142.77s user 40.95s system 343% cpu 53.452 total

### macOS M1 in asdf 
```
Executed in   24.23 secs    fish           external
   usr time   59.70 secs    0.07 millis   59.70 secs
   sys time    7.97 secs    1.13 millis    7.97 secs
```

(`time` output)

baileys-li:
> asdf has conflict with nvm.

AFAIK it's not asdf causing the problems, it's pnpm. pnpm is trying to be an alternative not only for npm, but also for [nvm](https://pnpm.io/cli/env) for whatever reason.

And if the console in the current folder shows that Node 22 is being used, pnpm is still using Node 20 under the hood.

At least this is the problem I had with Volta

mike-minchenko:
For the first time in my life, I decided to try to do something in the open source project, and decided to do everything as Andrei told to do, run through containers. The launch failed))) 😄

Mackbook M1

ai:
@mike-minchenko what error did you have? We may need a little tuning for Docker and Mac.

mike-minchenko:
Снимок экрана 2024-05-20 в 12 14 49
 
This was the first error. After which I added the line to --platform linux/amd64. Everything started, but all the RAM on my laptop was loaded and the computer began to freeze.

ai:
@mike-minchenko what way do you use to run Docker images?

ai:
@mike-minchenko I mean what tool do you have to run Docker images in the system?

mike-minchenko:
I'm using Docker Desktop for Apple Silicon. 

I can run other images.

I'm not very familiar with Docker and its settings. 
I'll look into it

PR #203, opened 2024-05-16
Do not open first candidate link on mobile

### LINKED ISSUE #198: Do not open first candidate link on mobile
Right now on mobile feed add UI looks weird. After entering URL you need to click on already pressed first candidate.

It happens because on desktop it is cool to open first candidate automatically, but on mobile not.

The idea:
1. Add `isMobile` to environment
2. Define `isMobile` in web environment to use media queries (the same as we use for 1 step per page mode)
3. Use `env.isMobile` in `core/preview.ts` and do not open candidate automatically on mobile
4. Add tests
5. Think about cleaning from some from https://github.com/hplush/slowreader/pull/186 like `toggleToSecondStep` argument

ilyatitovich:
Fixes #198 

- Added `isMobile` store in `core/environment.ts`
- Added logic that set `isMobile` on app loading and window resize in `web/main/environment.ts`
- Added condition that do not show candidate on mobile screen in `core/preview.ts`
- Added tests

# Motivation

This will improve the user experience on small screens.

# Checklist

- [x] Run `pnpm test`.
- [x] Think about testing
- If you changed web client:
- If you changed `core/`:

### DIFF
--- a/core/environment.ts
+++ b/core/environment.ts
@@ -120,6 +120,12 @@ export function getEnvironment(): Environment {
   return currentEnvironment
 }
 
+export const isMobile = atom<boolean | undefined>()
+
+export function setIsMobile(isSmallScreen: boolean): void {
+  isMobile.set(isSmallScreen)
+}
+
 const testRouter = atom<BaseRoute | undefined>()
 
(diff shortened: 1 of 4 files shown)


ai:
Nice! I like hos it was done.

PR #208, opened 2024-05-18
Add candidate to preview URL

### LINKED ISSUE #201: Add candidate to preview URL
We do not keep the current selected candidate in `/feeds/add` URL.

Let’s add it. It will simplify the `two-page` logic a lot since we will not have route-less back logic.

- [ ] Add `candidate` to `core/router.ts`
- [ ] Add route to `web/stores/router.ts`
- [ ] Change `core/preview.ts` to use router
- [ ] Change web client to use links instead of buttons on Add Feed page

ilyatitovich:
Fixes #201 

- Added `candidate` in `add` route and logic to create `backRoute` for `add` route
- Added logic to use router for setting and show candidate in `core/preview.ts`
- Added `candidate` in URl in `web/stores/router.ts`
- Updated web client to use links instead of buttons for set candidate and `Back` button in mobile navbar
- Added tests

I would like to discuss this. Maybe we can improve the code and logic.

# Checklist

- [x] Run `pnpm test`.
- [x] Think about testing
- If you changed web client:
- If you changed `core/`:


core/preview.ts
```diff
@@ -117,8 +116,14 @@ function addCandidate(url: string, candidate: PreviewCandidate): void {
 
   $links.setKey(url, { state: 'processed' })
   $candidates.set([...$candidates.get(), candidate])
-  if ($candidates.get().length === 1) {
-    if (!isMobile.get()) setPreviewCandidate(url)
+  if ($candidates.get().length === 1 && !isMobile.get()) {
+    getEnvironment().openRoute({
+      params: {
+        candidate: candidate.title,
```
ai, line 122: Why not URL?

ilyatitovich: For aesthetic reasons, so that the URL looks good in the browser, for example: `/add/https%3A%2F%2Fdev.to%2Ffeed%2F/DEV%2520Community`. But now I think it has potential problem if two candidates will have same titles. So should I use here `candidate.url`?

ai: Yes, I think URL is safer. Title can be empty or non uniwue.

ilyatitovich:
Changed 'title' to 'url' for candidate.

ai:
Try to reload page when some candidate is selected, you will get 404 (you can use staging or call `pnpm production` in `web/`).

But everything works without candidate in URL.

We use [this script](https://github.com/hplush/slowreader/blob/main/web/scripts/export-routes.ts) to generate `web/routes.regexp` and then we are [injecting](https://github.com/hplush/slowreader/blob/main/web/Dockerfile#L8) it into [nginx config](https://github.com/hplush/slowreader/blob/main/web/nginx.conf#L43).

ai:
> Updated pathRouter for handling page reload and avoiding 404 page.

But when I press Reload, the `candidate` valuer changes.

1. Open Add
2. Enter `github.blog`
3. Slow Reader will find 2 feeds: `Comments for …` and `GitHub Blog`. The Comments are selected by default.
4. Click on `GitHub Blog`
5. Reload the page.
6. The comments are selected, not the `GitHub Blog`.

And during the refresh I see how `?candidate=` is removed and set again.

ilyatitovich:
I think this happens because when we reload the page, `setPreviewUrl()` and other processes are ran. Maybe it’s worth providing a cache for the current candidate, for example, in `localStore`? We can update it when user type the different URL and clear when user moved to another route from `add` route.

ai:
> Maybe it’s worth providing a cache for the current candidate, for example, in localStore

No, user can open multiple tabs with different URL.

The better is to change `setPreviewURL()` logic. For instance, we can not to auto-set candidate if there is something different in router.

ai:
Thanks! It was much more complex task that we thought in the beginning.

PR #209, opened 2024-05-18
Check sha256 sum of the node binaries

nick318:


proxy/Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM registry.access.redhat.com/ubi9/ubi:9.4 as builder
 
 ENV NODE_VERSION=22.2.0
-RUN curl -fsSL https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.gz \
-  | tar -xz -C /usr/local/ --strip-components=1
+ADD --checksum=sha256:2c6eaf8bfd0f886ed8764ffce19c795e909639105e4056b1d9f8f917bad12cf1 https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.gz /usr/local
```
ai, line 4: Nice! Didn’t know that `ADD` accept URL

proxy/Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM registry.access.redhat.com/ubi9/ubi:9.4 as builder
```
ai, line 1: Can I ask you to do the same for `.devcontainer/Dockerfile`? It is the same `Node.js` binary. Just to make commit looks better.

nick318: @ai I made a change, but could you run it locally please, because I could not verify `.devcontainer/Dockerfile` on my machine (I suppose because of the arm arch)

ai:
Thanks! I improved it to pnpm and update script.

PR #211, opened 2024-05-19
Fixed small typo in CONTRIBUTING.md

uuladzislau:
I noticed a small typo on screenshot of `CONTRIBUTING.md` file shared in one of Andrei's recent tweets, and it bothered me so much, so I went ahead and submitted a PR to correct it.

I didn't run any tests because I am 100% sure this change won't break anything. However, please let me know if you would still like me to perform the standard set of tests and checks on the project.

PR #213, opened 2024-05-19
Fix table of content

uuladzislau:
It appears that PR https://github.com/hplush/slowreader/pull/211 has caused test failures. The issue stems from the title change in the PR, while the corresponding references in the document were not updated to match.


ai:
Thanks!

PR #217, opened 2024-05-26
Add visual tests for Slow/Fast page

### LINKED ISSUE #179: Add visual tests for Slow/Fast page
We need Storybook stories for Slow and Fast posts.

What cases we need to cover:
1. 3 posts on the page
2. Many posts so we have a pagination
3. Case 1, but some post is open

The problem is that we can’t use `lorem ipsum` and other random text.

We need some real posts. I suggest using [Wiki news](https://en.wikinews.org/wiki/Main_Page) (something neutral, like science news).

baileys-li:
Fixes #179

# Motivation

# Screenshot or Video

# Checklist

- [x] Run `pnpm test`.
- [x] Think about testing
- If you changed web client:
- If you changed `core/`:


baileys-li:
Maybe I need some consultation about app functionality and svelte 😅

core/lib/stores.ts
```diff
@@ -11,6 +11,11 @@ type StoreValues<Stores extends ReadableAtom[]> = {
   [Index in keyof Stores]: StoreValue<Stores[Index]>
 }
 
+/** use only for test mocks or don't permit rewrite store at all */
+export const isWritableAtom = <Value>(
```
ai, line 15: If you want to avoid `/* ts-expect-error */` let’s create some method like `forceSet<Value>(store: ReadonlyAtom<Value>, value: Value)`

Right now it is hard to understand why we are checking `isWritableAtom`, because other developer will think that sometimes store is writable, sometimes not.

web/stories/pages/posts/mocks.ts
```diff
@@ -0,0 +1,97 @@
...
...ing data can improve performance by avoiding repeated network requests.<br>\nPersisting user state or information across sessions.</p>\n\n<p>Session Storage: Temporary storage for data during a single session, such as form data that hasn't been submitted yet. Information that should be cleared when the user closes the tab, like temporary state information that is session-specific.</p>\n\n\n\n\n<h2>\n  \n  \n  Security Considerations\n</h2>\n\n<p>Local Storage: Since data persists indefinitely, it can be a target for attackers if sensitive information is stored without proper security measures. It is essential to avoid storing sensitive data like passwords or personal information in local storage.</p>\n\n<p>Session Storage: Data has a shorter lifespan, reducing the risk window for potential exploitation. However, it still should not be used for highly sensitive information without proper security measures.</p>\n\n\n\n\n<p>That's all!! </p>\n\n<p>This is all the stuff that I could find out. If you have any other thing in mind which I missed just go to comments and mention about the same!</p>\n\n<p>Stay happy, Keep hustling<br>\nSigning off Meet Bhalodiya,</p>\n\n<p>Peace ✌️</p>\n\n",
```
ai, line 18: The text is part of copyright law. We can’t just copy someone text from Internet without a permission.

This is why original issue suggested taking Wikinews, since it contains only copyleft text.

ai:
> Maybe I need some consultation about app functionality and svelte 😅

No, it is pretty solid PR.

We just need to fix a few small issues and we are ready to merge.

web/stories/pages/posts/slow.stories.svelte
```diff
@@ -0,0 +1,34 @@
+<script context="module" lang="ts">
+  import { Story } from '@storybook/addon-svelte-csf'
+
+  import { MOCKED_POSTS } from './mocks.js'
+  import MockedSlow from './slow-mock-env.svelte'
+
+  export const meta = {
+    component: MockedSlow,
+    title: 'Pages/Posts/Slow'
```
ai, line 9: Let’s have `Pages/Slow` (remove `Posts` step). Right now we group pages only if they have some common folder in URL (like `/feeds/add` or `/settings/profile`).

web/stories/pages/posts/slow-mock-env.svelte
```diff
@@ -0,0 +1,36 @@
...
+  export let post: PostValue | undefined = undefined
+
+  onMount(() => {
+    if (
+      isWritableAtom(slowPosts) &&
+      isWritableAtom(totalSlowPages) &&
+      isWritableAtom(totalSlowPosts) &&
+      isWritableAtom(openedSlowPost)
+    ) {
+      slowPosts.set(state)
```
ai, line 27: Can we move it to `scene.svelte`?

We will need to test `slowPosts` stores in other stories too later. Also, right now all stores mocking are in the same place, which is very flexible.

web/stories/pages/posts/mocks.ts
baileys-li: That's why I wanted to set up a consultation.

When I use the English channel, the app only shows one post from it. I don't use RSS readers, so I'm not sure how it should be used.

ai: > the app only shows one post from it

What app? I mean https://en.wikinews.org/wiki/Main_Page

web/stories/pages/posts/slow-mock-env.svelte
baileys-li: Sure

core/lib/stores.ts
baileys-li: Can I replace the readonlyExport function everywhere with a simple type casting?
I think function a little confusing, like it should delete method `.set()` from store, but it's just ts util

web/stories/pages/posts/mocks.ts
baileys-li: Slowreader.

Yeah, I used the [WikiNews Atom Feed](https://en.wikinews.org/w/index.php?title=Special:NewsFeed&feed=atom&categories=Published&notcategories=No%20publish%7CArchived%7CAutoArchived%7Cdisputed&count=50&hourcount=200&stablepages=only).
When I add this feed on setting, the app show manypost  previews, but on slow and fast page of the app, there is only one post.

For example "Веб-стандарты" loads all the post from feed.
I thinks that's because Web Standards doesn't have date. So there is somewhere in source code filter of old posts

core/lib/stores.ts
ai: > Can I replace the readonlyExport function everywhere with a simple type casting?

No. We use types to check that API is used in correct way.

Using `.set()` for stores with `readonlyExport` could break their logic. It is OK for stories (where we do only a snapshot), but will be bad in client or test.

web/stories/pages/posts/mocks.ts
ai: > When I add this feed on setting, the app show manypost previews, but on slow and fast page of the app, there is only one post.

It is not a bug, but how all RSS readers works.

RSS reader is to read new posts. When you subscribe to the feed, all posts are old. So we do not show them.

Think about newsletter. On the newsletter page you see all old letters. But when you put your email to form and subscribe, you will get new email only later.

(I will add an issue to improve UI and explain it to user.)

web/stories/pages/posts/mocks.ts
```diff
@@ -0,0 +1,185 @@
+import type { FastEntry, FeedValue, PostValue } from '@slowreader/core'
+
+const MOCKED_FEEDS: FeedValue[] = [
```
ai, line 3: The raw news from Wikinews is so ugly.

Each has this block:
Captura desde 2024-05-27 00-29-49

[Citizen journalist, Viktor Pinchuk — gives travel presentation in Sevastapol](https://en.wikinews.org/wiki/Citizen_journalist,_Viktor_Pinchuk_%E2%80%94_gives_travel_presentation_in_Sevastapol) that news has a huge block.

I understand that it is how Slow Reader see them right now. But here we are testing pagwe design, not Atom loader. Can you just cut this block and trancuate the second post?

ai: Also, it is not cool to bloat git with unnecessary big data :D

core/lib/stores.ts
baileys-li: No, types will remain.

I mean something like this
```diff
-export const slowPosts = readonlyExport($posts)
+export const slowPosts: ReadableAtom<SlowPostsValue> = $posts
```
or
```diff
-export const slowPosts = readonlyExport($posts)
+export const slowPosts: MakeReadOnly<typeof $posts> = $posts
```

Same typecasting. But without extra function in vanilla

web/stories/pages/posts/mocks.ts
baileys-li: Roger

core/lib/stores.ts
ai: It was too long, I added helper to make it shorter.

But I understand you, functions should change JS, not types. Good point.

Let’s keep `readonlyExport` for this PR, I will think about it.

web/stories/pages/posts/mocks.ts
baileys-li: I was not sure how 'real' real data should be¯\_(ツ)_/¯

Squash PR when it will be ready to merge pls

core/lib/stores.ts
```diff
@@ -11,9 +12,17 @@ type StoreValues<Stores extends ReadableAtom[]> = {
...
 }
 
-export function readonlyExport<Store extends ReadableAtom>(
-  store: Store
-): ReadableAtom<StoreValue<Store>> {
+/**
+ * use only for manual write store in tests
+ * otherwise don't permit rewrite store at all
+ **/
+export function forceSet<Value>(store: NanoStore<Value>, value: Value): void {
```
ai, line 19: I think wer can use `ReadableAtom` instead of `Store`/`NanoStore` here. At least it will allow avoid import renaming.

web/stories/pages/posts/mocks.ts
ai: > I will manually edit… lastPublishedAt in FeedValue I guess?

Yes

web/stories/pages/posts/fast-scene.svelte
```diff
@@ -0,0 +1,21 @@
+<script lang="ts">
```
ai, line 1: Can we avoid special `FastScene` and `SlowScene` and just use `Scene`?

We will need to set posts in many places.

baileys-li: I'm kinda afraid that `Scene` become God component.
Maybe it's better to keep `FastScene` and `SlowScene` but only contain a slot in them?

ai: The God component is sort of a plan here.

I understand that it is a bad practice for general component. But what is the problem for mocking component?

ai:
Thanks!

PR #220, opened 2024-05-28
Add workflow to write auto-review comment

pshergie:
No issue

# Checklist

- [ ] Run `pnpm test`.
- [ ] If you added a new dependency, check our [requirements](https://github.com/hplush/slowreader/blob/main/README.md#dependencies).
- [ ] Think about testing
- If you changed web client:
- If you changed `core/`:


.github/data/auto-review-comment.yml
```diff
@@ -0,0 +1,8 @@
+- prependMsg: "🗯️ [pull-request-auto-reviewer]:"
```
ai, line 1:
```suggestion
- prependMsg: "Pull request auto-reviewer"
```

Let’s do it less robotic :)

.github/workflows/auto-review-comment.yml
```diff
@@ -0,0 +1,34 @@
+name: Auto Review Comment
+
+on:
+  workflow_run:
+    workflows:
+      - Diff Compile
```
ai, line 6: Where do we have this worflow?

.github/workflows/auto-review-comment.yml
```diff
@@ -0,0 +1,34 @@
...
+
+permissions:
+  pull-requests: write
+
+jobs:
+  auto-review:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
```
ai, line 18:
```suggestion
      - name: Checkout the repository
```

Let’s use the same name

.github/workflows/auto-review-comment.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: actions/checkout@v4
+      - name: Download Diff Artifact
```
ai, line 20:
```suggestion
      - name: Download PR diff from auto review prepare step
```

.github/workflows/auto-review-comment.yml
pshergie: I am planning to add it in the next PR, it doesn't matter for now imo

.github/workflows/auto-review-comment.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: auto-review-artifacts
```
ai, line 23: We do not use `artifacts` postfix in other artifacts. Let’s use `pr-diff`

.github/workflows/auto-review-comment.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  workflow_run:
+    workflows:
+      - Diff Compile
+    types:
+      - completed
+
+permissions:
+  pull-requests: write
+
```
ai, line 12: We don’t have new lines in other workflows

.github/workflows/auto-review-comment.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Download Diff Artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: auto-review-artifacts
+          path: auto_review_artifacts/
+          repository: ${{ github.repository_owner }}/${{ github.event.repository.name }}
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
ai, line 27: Are you sure that it will work with standard `GITHUB_TOKEN` and we do not need special token?

.github/workflows/auto-review-comment.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Analyze changes
+        uses: ./
+        id: auto-review-action
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          datapath: .github/data/auto-review-comment.yml
```
ai, line 34: Let’s move to `.github/auto-review-comment.yml`

.github/workflows/auto-review-comment.yml
ai: Got it. Let’s rename if to `Auto-review Prepare` to match other prepare-do workflows.

ai: Or `Auto-review Diff Prepare`

PR #225, opened 2024-06-04
Add desktop support for going back from the steps

### LINKED ISSUE #197: Add desktop support for going back from the steps
We have `<TwoStepPage>` and the logic to move back in those steps [for mobile](https://github.com/hplush/slowreader/pull/186).

We need to add the same back logic to desktop cases:

1. We need to add some header to second step with `×` button. We should not show this header on mobile.
2. We should check that `Escape` works as expected and also closes second step. We need to be sure, that it doesn’t have conflict with `keyux` jumps.
   1. We need to test all 3 cases: Slow posts, Feed add, Feed categories
   2. We need to test Esc after keyboard interactions (hotkeys, arrows, Enter) and after the mouse clicks

ilyatitovich:
Fixes #197 

I would like to discuss it.

- Added `Back` button for desktop view. The button is wide. Should it be square?
- Added `Escape` key handler. It’s strict with `activeElement` checking so, if I right understand [keyux jump](https://github.com/ai/keyux/blob/main/jump.js#L5) only working if current element is no `<body>`. Now I don't see any conflict.  
- Updated dependency for UI elements from `secondStep` to `backRoute`. Since we have all logic for creating `backRoute`, maybe it’s not necessary to use `two-steps.ts` and we can delete it?

# Checklist

- [x] Run `pnpm test`.
- [x] Think about testing
- If you changed web client:


web/ui/two-steps-page.svelte
```diff
@@ -23,25 +27,38 @@
...
   >
     <slot name="one" />
   </div>
   <div
     bind:this={second}
-    class={`two-steps-page_step ${!$secondStep ? 'is-hidden' : ''}`}
+    class={`two-steps-page_step ${!$backRoute ? 'is-hidden' : ''}`}
   >
+    {#if $backRoute}
+      <div class="two-steps-page_close-button">
```
ai, line 58: Let’s add `<aside>` with `<Button secondary hiddenLabel>` to make it square

core/router.ts
```diff
@@ -214,6 +214,12 @@ export function removeFeedFromRoute(): void {
   }
 }
 
+export function backToFirstStep(): void {
+  if (backRoute.get()) {
```
ai, line 218: You can fix TS error by:

```ts
let back = backRoute.get()
if (back) {
  getEnvironment().openRoute(back)
}
```

core/test/two-steps.test.ts
```diff
@@ -126,3 +128,22 @@ test('works with slow route', async () => {
     route: 'slow'
   })
 })
+
+test('back to first step', async () => {
```
ai, line 132: The verb `back` for `he/she/it` should have `s` on the end.

We are using RSpec naming system.

```suggestion
test('goes back to first step', async () => {
```

core/test/two-steps.test.ts
```diff
@@ -126,3 +128,22 @@ test('works with slow route', async () => {
...
+    params: { feed },
+    route: 'categories'
+  })
+
+  backToFirstStep()
+
+  deepStrictEqual(router.get(), {
+    params: {},
+    route: 'categories'
+  })
```
ai, line 148: Let’s make test more tricky and call `backToFirstStep()` one more in the end

web/ui/two-steps-page.svelte
```diff
@@ -23,25 +27,38 @@
     })
   }
 
+  function handleEscapeKey(event: KeyboardEvent): void {
```
ai, line 30: The current flow works only if you use mouse before only then press `Esc`.

But if you use keyboard, it will not work as expected.

Try this flow:
1. Press <kbd>m</kbd>
2. Press <kbd>↵</kbd> (to open Add feed)
3. Type `github.blog`
4. Press <kbd>↵</kbd> (to move to found items)
5. Press <kbd>↵</kbd> (to move to side panel)
6. Press <kbd>Esc</kbd>. Expected: move back to search form and close side panel. Actual behavior: panel is still open.

ilyatitovich:
Updated `handleEscapeKey()` function.

1. Now we use **capturing phase** in `addEventListener` that allow us to stopping propagation early in the event flow.
2. Then we check route and if it `fast` or `slow` we stop propagation for prevent keyux jump, because for `add` and `categories` routes is different flow and we need jump back to list + close panel, but for `slow` and `fast` routes we need just close panel.

web/ui/two-steps-page.svelte
```diff
@@ -23,25 +27,42 @@
...
+    if (backRoute.get() && event.key === 'Escape') {
+      let { route } = router.get()
+      if (route === 'fast' || route === 'slow') {
+        event.stopPropagation()
```
ai, line 34: This line block `keyux` behavior, which reduce UX.

Here is use case:
1. Open `Add feed`
2. Enter `github.blog`
3. Press Enter (go to candidates list)
4. Press Enter (go to candidate sidebar)
5. Press Esc. Expected: sidebar is closed, focus is back in candidate list. Actual: sidebar is closed but no element is in the focus.

ai: Why do we need to disable `keyux` at all?

ilyatitovich: I suggest using blocking `keyux` jump back only on the `fast` and `slow` routes. When the user focuses on the `Open post` button and presses `Enter`, the focus remains on the `Open post` button. Therefore, when the user closes the side panel with `Escape`, we don't need to jump back to the categories list but should keep the `Open post` button in focus.

This behavior can be seen here: [video](https://drive.google.com/file/d/1rXhIkkSQ0cTVAuubfS5Juby-cbGuxXIJ/view?usp=sharing).

However, on the `add` and `categories` routes, we still use the `keyux` jump back when `Escape` is pressed.

This behavior can be seen here: [video](https://drive.google.com/file/d/1yDopW2odQ2-RGy4KWbpyrrbNLtI11JAl/view?usp=drive_link).

ai: > I suggest using blocking keyux jump back only on the fast and slow routes

fast/slow just has not `keyux` integration yet, so don’t look on them.

When I will have final design for these page, I will add special `aria` and focus will jump from post card to post side panel (like in Feed Add).

(Sorry, I forgot to mention it in the task 😅)


> However, on the add and categories routes, we still use the keyux jump back when Escape is pressed.
> This behavior can be seen here: [video](https://drive.google.com/file/d/1yDopW2odQ2-RGy4KWbpyrrbNLtI11JAl/view?usp=drive_link).

This video looks great. Just double-check that everything works on mouse and in the mouse/keyboard mixed scenarios.

If yes, let’s just keep `keyux` logic and just add a logic to close sidepanel.

ilyatitovich:
I changed the conditions for stopping event propagation:

1. Added `role` checking - this allows us to avoid unexpected jumps to the side menu if the user clicks on a list item with the mouse and then presses `Escape`. Now, in this case, the focus will move to the clicked list item.
2. Added `tagName` checking to prevent jumping back if the user clicks on a list item, then clicks somewhere else on the screen, and then presses `Escape` to close the side panel.

Test: [video](https://drive.google.com/file/d/1J8gKo1ArIc9E3EguK7fyEaCJkymlMqu2/view?usp=sharing)

ai:
Works amazing. Thanks for helping me polish this.

Hope it helps you to go deeper into keyboard UX world (feel free to ask questions about that `role`s)

ilyatitovich:
Yes, I learned a lot :D

PR #227, opened 2024-06-07
Add request rate limit to proxy

### LINKED ISSUE #168: Add request rate limit to proxy
**Blocked by #100**

To prevent abusing our public CORS proxy

- [ ] Allow only 500 request in a minute to the same domain from the same IP-address.
- [ ] Allow only 5000 request to any domain from the same IP-address.

On passing those limits:
1. Print to console (we will need it to debug)
2. Do not cancel response, just postpone (so client will not see an error, just slower connection).

These changes need to be done in `proxy/`. It will be cool to have tests, but if they are not stable or hard to make, we can avoid it.

janefawkes:
Fixes #168 

I sort of tested this by running the proxy and spamming it with batches of requests (like [this](https://gist.github.com/janefawkes/b3cf23c02d7295fc96cd31d9f214718c)). It works on my machine™, but `pnpm test` doesn't quite agree with it. 

Would love to hear any sort of feedback on this, because I'm not sure if I'm moving in the right direction or not.

Additional changes: proxy config and error handling have been divorced from the `createProxyServer` implementation.

### DIFF
--- a/proxy/package.json
+++ b/proxy/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "start": "tsx watch index.ts",
-    "test": "FORCE_COLOR=1 pnpm run /^test:/",
+    "test": "pnpm run /^test:/",
     "build": "esbuild index.ts --bundle --platform=node --sourcemap --format=esm --outfile=dist/index.mjs",
     "production": "node --run build && ./scripts/run-image.sh",
     "test:proxy-coverage": "c8 bnt",
--- a/proxy/proxy.ts
+++ b/proxy/proxy.ts
@@ -57,7 +57,7 @@ export function isRateLimited(
     return true
   }
 
-  rateLimitInfo.count += 1
+  rateLimitInfo.count++
   store.set(key, rateLimitInfo)
 
   return false
@@ -72,6 +72,10 @@ export function checkRateLimit(ip: string, domain: string): boolean {
 }
 
 export function handleError(e: unknown, res: ServerResponse): void {
+  if (res.headersSent) {
+    // Headers already sent, cannot handle error
+    return
+  }
   // Known errors
   if (e instanceof Error && e.name === 'TimeoutError') {
     res.writeHead(400, { 'Content-Type': 'text/plain' })
--- a/proxy/proxy.ts
+++ b/proxy/proxy.ts
@@ -40,45 +40,35 @@ const RATE_LIMIT = {
 let rateLimitMap: Map<string, RateLimitInfo> = new Map()
 let requestQueue: Map<string, Promise<void>> = new Map()
 
-function isRateLimited(ip: string, domain: string): boolean {
+function isRateLimited(
+  key: string,
+  store: Map<string, RateLimitInfo>,
+  limit: { LIMIT: number; DURATION: number }
+): boolean {
   let now = performance.now()
+  let rateLimitInfo = store.get(key) || { count: 0, timestamp: now }
 
-  let domainKey = `${ip}:${domain}`
-  let domainRateLimit = rateLimitMap.get(domainKey) || {
-    count: 0,
-    timestamp: now
+  if (now - rateLimitInfo.timestamp > limit.DURATION) {
+    rateLimitInfo.count = 0
+    rateLimitInfo.timestamp = now
   }
 
-  if (now - domainRateLimit.timestamp > RATE_LIMIT.PER_DOMAIN.DURATION) {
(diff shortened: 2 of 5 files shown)


proxy/proxy.ts
```diff
@@ -13,17 +13,179 @@ class BadRequestError extends Error {
...
+    LIMIT: 500,
+    DURATION: 60 * 1000
+  },
+  GLOBAL: {
+    LIMIT: 5000,
+    DURATION: 60 * 1000
+  }
+}
+
+const delay = (ms: number): Promise<void> => {
```
ai, line 39: Yoou can use:

```
import { setTimeout } from 'node:timers/promises`

await setTimeout(50)
```

proxy/proxy.ts
```diff
@@ -13,17 +13,179 @@ class BadRequestError extends Error {
...
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
+
+const rateLimitMap: Map<string, RateLimitInfo> = new Map()
+const requestQueue: Map<string, Promise<void>> = new Map()
+
+const isRateLimited = (ip: string, domain: string): boolean => {
```
ai, line 46: We prefer `function isRateLimit() {}` instead of `const isRateLimit = () => {}`

proxy/proxy.ts
```diff
@@ -13,17 +13,179 @@ class BadRequestError extends Error {
...
+    LIMIT: 5000,
+    DURATION: 60 * 1000
+  }
+}
+
+const delay = (ms: number): Promise<void> => {
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
+
+const rateLimitMap: Map<string, RateLimitInfo> = new Map()
```
ai, line 43: All variables which is not `NAMED_LIKE_THIS` should be `let` according to our code style

proxy/proxy.ts
```diff
@@ -13,17 +13,179 @@ class BadRequestError extends Error {
...
+
+  domainRateLimit.count += 1
+  globalRateLimit.count += 1
+  rateLimitMap.set(domainKey, domainRateLimit)
+  rateLimitMap.set(globalKey, globalRateLimit)
+
+  return false
+}
+
+const handleError = (e: any, res: ServerResponse): void => {
```
ai, line 87:
```suggestion
const handleError = (e: unknown, res: ServerResponse): void => {
```

proxy/proxy.ts
janefawkes: > All variables which is not `NAMED_LIKE_THIS` should be `let` according to our code style

So I should not acknowledge constant references and use `const` only for real constants?

ai: Yes, use `let` for variables (ESLint will tell you some cases when you run it)

proxy/proxy.ts
janefawkes: I guess [that](https://github.com/hplush/slowreader/pull/227/commits/729ee25fa33adede8dea6bcdeb110172d4bb6973) should do the job

proxy/proxy.ts
```diff
@@ -13,16 +14,178 @@ class BadRequestError extends Error {
...
+    domainRateLimit.count = 0
+    domainRateLimit.timestamp = now
+  }
+
+  if (domainRateLimit.count >= RATE_LIMIT.PER_DOMAIN.LIMIT) {
+    return true
+  }
+
+  let globalKey = ip
+  let globalRateLimit = rateLimitMap.get(globalKey) || {
```
ai, line 62: The logic is same between `globalKey` and `domainKey`.

Maybe we should have function like:

```
function isRateLimited(key: string, store: Map<string, RateLimitInfo>, limit: { LIMIT: number, DURATION: number })
```

ai:
Now you need to fix tests (it is better to add some test for rate limit too) and we are ready to merge.

You can run tests by `cd proxy` and `pnpm test`.

You also need to fix ESLint warnings by calling `eslint --fix .`

janefawkes:
ESLint warnings are fixed.

Not sure what to do with tests, though. I am a bit stuck with them and I don't really know how to ensure proper coverage
image

proxy/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "start": "tsx watch index.ts",
-    "test": "FORCE_COLOR=1 pnpm run /^test:/",
```
janefawkes, line 11: Changed it for the debug purposes, gonna add this line back in the next commit.

PR #228, opened 2024-06-11
auto-review-comment: watch different file change

pshergie:
A better approach to handle changes in dependencies as `package.json` file can be updated without any dependency introduction.

# Checklist

- [ ] Run `pnpm test`.
- [ ] Think about testing
- If you changed web client:


pshergie:
the action shouldn't have posted the message regarding dependencies change, let me investigate it

pshergie:
Closed as a similar changes has been already added to the main branch

PR #234, opened 2024-07-30
Adds Slowreader Chrome extension

ninaTorgunakova:
Adds Chrome extension which allows Slowreader to fetch data from any site without using a proxy server.


extension/.eslintrc.cjs
```diff
@@ -0,0 +1,48 @@
+module.exports = {
```
ai, line 1: Can you remove this?

extension/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
```
ai, line 1: Can you remove this?

extension/lefthook.yml
```diff
@@ -0,0 +1,6 @@
+pre-commit:
```
ai, line 1: We have own pre-commit config
https://github.com/hplush/slowreader/tree/main/.husky
https://github.com/hplush/slowreader/blob/main/nano-staged.json

Can you remove this?

extension/pnpm-lock.yaml
```diff
@@ -0,0 +1,3695 @@
+lockfileVersion: '9.0'
```
ai, line 1: Can you remove lock file since we are using pnpm monorepo with one lock file for every subdir?

extension/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
```
ai, line 1: We have own `tsconfig.json` in the root. Can you remove it?

extension/build.js
```diff
@@ -0,0 +1,24 @@
+import fs from 'fs'
+import path from 'path'
+import { fileURLToPath } from 'url'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
```
ai, line 6: BTW, you don’t need this hack anymore. Just use `import.meta.dirname`

ninaTorgunakova: @ai thanks, that's useful! I changed

PR #235, opened 2024-08-06
Add workspaceFolder to comply with devcontainer spec

3timeslazy:
[DevContainer specification says](https://containers.dev/implementors/json_reference/#image-specific) that `workspaceMount` and `workspaceFolder` must be set together. While the VS Code Extension works fine without `workspaceFolder` , it's not the case for DevPod. That PR fixes that.


ai:
Thanks! Respect for using Podman too :)

PR #248, opened 2024-09-12
Use glob in /scripts

### LINKED ISSUE #239: Move scripts to Node.js `glob`
In many scripts of `scripts/` and `web/scripts/` we use a lot of JS code to walk through the folders.

We can make code simpler by moving to glob. In Node.js there is native glob so we don’t need library for that.

3timeslazy:
Fixes #239

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -61,9 +61,12 @@ export default [
   {
     files: ['**/scripts/**.ts'],
     rules: {
-      'n/no-unsupported-features/node-builtins': ['error', {
+      'n/no-unsupported-features/node-builtins': [
+        'error',
+        {
           ignores: ['fs.globSync']
-      }]
+        }
+      ]
     }
   }
 ]
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -18,7 +18,7 @@ export default [
     }
   },
   {
-    files: ['web/pages/**/*', 'web/ui/**/*', 'web/main/**/*', 'core/test/**/*', 'scripts/*'],
+    files: ['web/pages/**/*', 'web/ui/**/*', 'web/main/**/*', 'core/test/**/*'],
     rules: {
       'n/no-unsupported-features/node-builtins': 'off'
     }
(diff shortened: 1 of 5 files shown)


ai:
CI is broken to fix it:

1. Run `eslint --fix .` to sort variables
2. Add `fs.globSync` to `ignore` of `n/no-unsupported-features/node-builtins` rule in ESLint config

scripts/check-versions.ts
```diff
@@ -71,14 +71,8 @@ if (packageManager !== pnpmFull) {
...
-    let relative = join(project, 'package.json')
-    checkPackage(relative, read(relative))
-  }
-  if (existsSync(join(ROOT, project, 'Dockerfile'))) {
-    let docker = join(project, 'Dockerfile')
-    checkDockerfile(docker, read(docker))
-  }
+let files = globSync(['./*/package.json', '**/Dockerfile'])
+for (let file of files) {
+  if (file.endsWith('package.json')) checkPackage(file, read(file))
```
ai, line 76: Let’s do 2 `globSync` calls here like

```js
for (let file of globSync('./*/package.json')) {
  checkPackage(file, read(file))
}
```

eslint.config.js
```diff
@@ -18,7 +18,7 @@ export default [
     }
   },
   {
-    files: ['web/pages/**/*', 'web/ui/**/*', 'web/main/**/*', 'core/test/**/*'],
+    files: ['web/pages/**/*', 'web/ui/**/*', 'web/main/**/*', 'core/test/**/*', 'scripts/*'],
```
ai, line 21: We disable this rule only for code running browser.

We can disable it for the whole project or make an allow-list of experimental API for Node.js related code like scripts.

3timeslazy: Would that work?

```js
{
    files: ['scripts/**.ts'],
    rules: {
      'n/no-unsupported-features/node-builtins': ['error', {
          ignores: ['fs.globSync']
      }]
    }
  }
```

3timeslazy:
@ai btw, it seems like `check-version` does not check Dockerfile versions. It matches against `node: (\d+\.\d+\.\d+)` pattern while Dockerfiles have `NODE_VERSION (\d+\.\d+\.\d+)` pattern

ai:
Yes, this is issue.

Can you fix it here if in separated PR?

3timeslazy:
> Yes, this is issue.
> 
> Can you fix it here if in separated PR?

I'll fix it in a separate PR than

scripts/check-versions.ts
3timeslazy: Done

eslint.config.js
3timeslazy: Done

ai:
Thanks!

PR #261, opened 2024-10-02
Do not show mobile menu on no categories

### LINKED ISSUE #178: Do not show mobile menu on no categories
On mobile, we show posts list after clicking on Fast/Slow buttons.

We need to close that menu if the list is empty of contains only 1 element (like 1 category for Fast or 1 feed for Slow).

Note, that Fast/Slow menus is loading the data and is not ready when user click on the menu. So it is better to use `core` API to check the categories. You may create some helpers (so we will re-use them in mobile client).

ilyatitovich:
Fixes #178  

1. Added API in `core` for display mobile menu;
2. Code handles open/close menu moved to `core` from template;
3. Added tests.

# Motivation
Improve UX.

I suggest to open menu when user navigate to `/slow`, `/fast` and `/add` routes, so I added subscription to `router`.  Also we can add check is mobile and handle calculation only if it's narrow screen. I would like to discuss it.


core/mobile-menu.ts
```diff
@@ -0,0 +1,47 @@
+import { atom, computed } from 'nanostores'
+
+import { fastCategories } from './fast.ts'
+import { router } from './router.ts'
+import { slowCategories } from './slow.ts'
+
+const isMenuShowed = atom<boolean>(false)
```
ai, line 7:
```suggestion
let isMenuShowed = atom<boolean>(false)
```

core/mobile-menu.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const isMenuHasItems = computed(
```
ai, line 9:
```suggestion
let isMenuHasItems = computed(
```

core/mobile-menu.ts
```diff
@@ -0,0 +1,47 @@
...
+
+router.subscribe((newRoute, prevRoute) => {
+  let targetRoutes = ['add', 'slow', 'fast']
+
+  if (
+    prevRoute &&
+    targetRoutes.includes(newRoute.route) &&
+    newRoute.route !== prevRoute.route
+  ) {
+    setTimeout(() => {
```
ai, line 43: Why do we have `setTimeout` here?

core/mobile-menu.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const isMenuOpened = computed(
+  [isMenuShowed, isMenuHasItems],
+  (isShowed, isHasItems) => {
+    return isShowed && isHasItems
+  }
+)
+
+router.subscribe((newRoute, prevRoute) => {
+  let targetRoutes = ['add', 'slow', 'fast']
```
ai, line 36: Let’s do a micro-optimization here by moving it to the root (we don’t need to define new set every time) and use object or `Set`

core/mobile-menu.ts
ai, line 7: The names are a littler confusing. I can’t understand why do we need this store.

ai, line 7: For internals store we use `$` prefix. Can we do it here too?

core/mobile-menu.ts
ilyatitovich: We have a listener for a click to close the menu on any part of the document, this is a micro delay before opening the menu on another route. When I test manually, sometimes the menu does not open when switching routes. But it works with a micro delay.

core/mobile-menu.ts
ilyatitovich: Could you explain me why using `let` better then `const`?

ai: Unification. You don’t need to switch between `const` and `let` if you need to change variable later.

Also, I came from Ruby where there is no such separation and use to have only one variable type.

web/ui/navbar/index.svelte
```diff
@@ -23,29 +25,23 @@
...
-  let isMenuOpened = $state(false)
-
-  function closeOnAnyClick(): void {
-    if (isMenuOpened) {
-      isMenuOpened = false
-      document.removeEventListener('click', closeOnAnyClick)
+  isMenuOpened.listen(isOpened => {
+    if (isOpened) {
+      setTimeout(() => {
+        document.addEventListener('click', toggleMenu)
```
ai, line 31: We should use here `closeMenu`. Toggle means that sometimes menu can be opened by click. It will be more clear to have explicit `closeMenu` and `openMenu` function for such cases.

core/mobile-menu.ts
ilyatitovich: I agree, I thought about how to name it and left it like this for now. The idea is that we use this store as a switch for opening and closing the menu, but we also use a filter so as not to open the menu on `/fast` and `/slow` routes if the menu does not have more than 1 category. Maybe name this store `isMenuOpened`, and the combined one `isMenuDisplayed`?

```js

export const isMenuDisplayed = computed(
  [$isMenuOpened, $isMenuHasItems],
  (isOpened, isHasItems) => {
    return isOpened && isHasItems
  }
)

```

ai: `Displayed` is a little better.

But maybe we can check `fast` categories count in `openMenu`? I think it will be simpler and less confusing.

ilyatitovich: I need to think about this, because I get some problems if I don't use the store. 

Also I suggest converting `/slow` and `/fast` links into buttons, dynamically removing the link from the `href` attribute, because if the user has a list of categories and wants to open the menu, he always goes to the `general` category. For example:
```js
href={$router.route === 'slow' ? "" : getURL('slow')}
```

ai: We need to be able to open link in new tab, this is why we can't convert it to button.

core/mobile-menu.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+  } else if (router.get().route === 'fast') {
+    let fastCategoriesValue = fastCategories.get()
+    if (
+      !fastCategoriesValue.isLoading &&
+      fastCategoriesValue.categories.length > 1
+    ) {
+      isMenuOpened.set(true)
+    }
+  } else if (!targetRoutes.has(router.get().route)) {
```
ai, line 25: Why do you need this check if `targetRoutes` contains only `slow` and `fast` and you already checked them above? We can use just `} else {`

web/ui/navbar/index.svelte
```diff
@@ -91,7 +88,7 @@
         aria-haspopup="menu"
         aria-keyshortcuts="u"
         href={getURL('slow')}
-        onclick={openMenu}
+        onclick={$router.route === 'slow' ? openMenu : () => {}}
```
ai, line 91: I think we should not disable click. Users may need it to go to another feed.

ai:
I am planning to start refactoring, so I am merging it (I will fix small changes)

PR #269, opened 2024-10-20
Check encode type for doanwload task

### LINKED ISSUE #268: Unpopular encoding
I don't think this is critical, I just wanted to notify about it. At the moment, slowreader might incorrectly display data received from websites that use less common encodings, such as Windows-1251 or ISO-8859-1. The percentage of such websites is very small, so feel free to close the issue if it's not relevant.

DisaPadla:
Fixes #268

### DIFF
--- a/core/download.ts
+++ b/core/download.ts
@@ -19,7 +19,7 @@ export interface DownloadTask {
   text(...args: Parameters<typeof request>): Promise<TextResponse>
 }
 
-enum EncodeType {
+export enum EncodeType {
   EUC_JP = 'euc-jp',
   ISO_2022_JP = 'iso-2022-jp',
   ISO_8859_1 = 'iso-8859-1',
--- a/core/download.ts
+++ b/core/download.ts
@@ -19,6 +19,18 @@ export interface DownloadTask {
   text(...args: Parameters<typeof request>): Promise<TextResponse>
 }
 
+enum EncodeType {
+  EUC_JP = 'euc-jp',
+  ISO_2022_JP = 'iso-2022-jp',
+  ISO_8859_1 = 'iso-8859-1',
+  SHIFT_JIS = 'shift-jis',
+  UTF8 = 'utf-8',
+  WINDOWS_1251 = 'windows-1251',
+  WINDOWS_1252 = 'windows-1252',
+  WINDOWS_1253 = 'windows-1253',
+  WINDOWS_31J = 'Windows-31J'
+}
+
 function detectType(text: string): string | undefined {
   let lower = text.toLowerCase()
   let beginning = lower.slice(0, 100)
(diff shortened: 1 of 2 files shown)


core/download.ts
```diff
@@ -40,6 +52,30 @@ function detectType(text: string): string | undefined {
   }
 }
 
+function detectEncodeType(response: Partial<Response>): string {
```
ai, line 55: I am hot sure that allow-list is the best option here since we will be limited to this list of encodings.

Maybe it is better to parse enconding from `headers.get('content-type')` string?

Like `/;\s*charset=(\w+)`

DisaPadla: thx. updated

core/test/download.test.ts
```diff
@@ -124,6 +124,22 @@ test('can download text by keeping eyes on abort signal', async () => {
   await rejects(response2, (e: Error) => e.name === 'AbortError')
 })
 
+test('detect unpopular encode type', async () => {
+  let task = createDownloadTask()
+
+  expectRequest('https://example.com').andRespond(
+    200,
+    'Hi',
```
ai, line 132: This test will pass even if we haven’t encoding feature.

What do you think if we will use CP1251 and put that broken (when CP1251 was parsed as UTF-8) encoding text here?

core/download.ts
```diff
@@ -40,6 +40,12 @@ function detectType(text: string): string | undefined {
...
```
ai, line 43:
```suggestion
function parseEncodeType(response: Partial<Response>): string {
```

Since we do not detect it (by byte orders, etc).

core/download.ts
```diff
@@ -40,6 +40,12 @@ function detectType(text: string): string | undefined {
...
+  let headers = response.headers ?? new Headers()
+  let contentType = headers.get('content-type')?.toLowerCase() ?? ''
+  return contentType.match(/charset=([a-zA-Z0-9-]+)/)?.[1] ?? 'utf-8'
```
ai, line 46:
```suggestion
  return contentType.match(/charset=(\w+)/)?.[1] ?? 'utf-8'
```

Can we do like this?

core/test/download.test.ts
DisaPadla: i will check later

core/download.ts
DisaPadla: i think, no. in this case we get `utf` instead of `utf-8`

core/download.ts
DisaPadla: updated

core/download.ts
ai: You right. What about `[\w-]+`?

ai: We also need `_`

Here is the full list of variants
https://www.iana.org/assignments/character-sets/character-sets.xhtml

DisaPadla: maybe better include all symbols except space? something like that?
`.match(/charset=([^\s]+)/)`

ai: `.match(/;\s*charset=([^\s;]+)/)`

Technically it could be like `Content-Type: text/html;charset=utf-8;boundary=ExampleBoundaryString`

core/test/download.test.ts
DisaPadla: hm, if i just cope paste broken text (like �), it doesnt work correctly. Continue research

core/download.ts
DisaPadla: oh, you are right. thx

core/test/download.test.ts
ai: Try binary format for symbol

ai: Another way is to get JS code for `UTF-8→CP1251` convertation and put:

```js
to1251('тест')
```

DisaPadla: sorry, what do you mean "Try binary format for symbol"?

ai: > sorry, what do you mean "Try binary format for symbol"?

[String.fromCodePoint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint)

PR #287, opened 2025-01-30
squash svelte each_key_duplicate error if same feed fails in multiple categories

alsuren:
Just squashes a svelte error each_key_duplicate error by deduplicating `$unLoadedFeeds`

# Motivation

I saw that there was a talk about this at FOSDEM, so I thought I'd check it out.

I was importing my feedly feeds, and I noticed that there was an error in the console. This is because I have the same url listed under two different categories. The app tries to import it twice and it fails twice, so the following #each raises a each_key_duplicate error.

```
      <ul class="feeds-import_unloaded">
        {#each $importUnLoadedFeeds as feed (feed)}
          <li>
            {feed}
          </li>
        {/each}
      </ul>
```

If you would rather I push the error out into the rendering code, or somehow deduplicate urls before I make the requests, I can have a stab at this instead.

See you at FOSDEM. :D

### DIFF
--- a/core/import.ts
+++ b/core/import.ts
@@ -10,6 +10,7 @@ import { addFeed, type FeedValue } from './feed.ts'
 import { readonlyExport } from './lib/stores.ts'
 import { importMessages } from './messages/index.ts'
 import { createFeedFromUrl } from './preview.ts'
+import { unique } from './loader/utils.ts'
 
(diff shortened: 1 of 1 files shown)


ai:
Wow! Thanks a lot!

PR #301, opened 2025-06-03
Add Link header support

### LINKED ISSUE #27: Add `Link` header support
In RSS and Atom (maybe even some universal helper) we need to be sure that link to feed wil be not in `<link>` tag, but in `Link` HTTP header

ashlkv:
Fixes #27

- Feed loaders now extract feed urls of their respective types from `Link` header using the shared utility function
- Added and updated the tests

# Motivation

The `Link` http header can sometimes contain the link to the feed. This http header is used when the document itself is difficult to modify.

- Feed loaders look for a feed url of their respective type in the `Link` http header
- If found, this link or links take priority over links found in the document's html
- Works even if the document is not html (should it?)

### DIFF
--- a/core/loader/utils.ts
+++ b/core/loader/utils.ts
@@ -76,7 +76,7 @@ export function findHeaderLinks(
   if (!linkHeader) {
     return []
   }
-  return linkHeader.split(/,\s?/).reduce((urls: string[], link: string) => {
+  return linkHeader.split(/,\s?/).reduce<string[]>((urls, link) => {
     let [, url] = link.match(/<(.*)>/) || []
     let attributes = link.split(/;\s?/)
     let matchesType = attributes.includes(`type="${type}"`)
--- a/core/loader/utils.ts
+++ b/core/loader/utils.ts
@@ -30,7 +30,7 @@ function buildFullURL(
   )
 }
 
-export function findLinksByType(text: TextResponse, type: string): string[] {
+export function findDocumentLinks(text: TextResponse, type: string): string[] {
   let document = text.parseXml()
   if (!document) return []
   return [...document.querySelectorAll('link')]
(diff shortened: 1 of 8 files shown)


core/loader/utils.ts
```diff
@@ -61,6 +61,36 @@ export function findAnchorHrefs(
     .map(a => buildFullURL(a, text.url))
 }
 
+/**
```
ai, line 64: Very good comment. But we don’t use it in another functions.

Can we keep the old comment-less format in PR?

But I am open to discuss adding more comments (just prefer to have some policy).

core/loader/utils.ts
```diff
@@ -30,7 +30,7 @@ function buildFullURL(
   )
```
ai, line 30: Can you run `pnpm format`? It will fix CI warning for code style.

core/loader/utils.ts
```diff
@@ -61,6 +61,36 @@ export function findAnchorHrefs(
...
+ */
+export function findHeaderLinks(
+  response: TextResponse,
+  type: string
+): string[] {
+  let linkHeader = response.headers.get('Link')
+  if (!linkHeader) {
+    return []
+  }
+  return linkHeader.split(/,\s?/).reduce((urls: string[], link: string) => {
```
ai, line 79:
```suggestion
  return linkHeader.split(/,\s?/).reduce<string[]>((urls, link) => {
```

By using type template we can avoid casting types for callback arguments.

ashlkv: Roger that

core/test/loader/utils.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { createTextResponse } from '../../download.js'
+import { findHeaderLinks } from '../../loader/utils.js'
+
+test('returns urls from link http header', () => {
+  deepStrictEqual(
+    findHeaderLinks(
+      createTextResponse(`<!DOCTYPE html><html><head></head></html>`, {
+        headers: new Headers({
+          Link: '<https://one.example.com>; rel="alternate"; type="application/rss+xml", <https://two.example.com>; rel="alternate"; type="application/rss+xml"'
```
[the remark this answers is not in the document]
ashlkv: Yep split is better

PR #302, opened 2025-06-08
Comments for feed loader and feed add

ashlkv:
Adds comments:

- to loader utilities
- to loader interface
- to feed add page
- to some of the not so obvious functions

# Motivation

It's sometimes difficult to tell what functions do. Especially for newcomers and especially now with no ui. We reconsidered our "no comments" policy and decided to start adding some.

### DIFF
--- a/core/current-page.ts
+++ b/core/current-page.ts
@@ -5,7 +5,7 @@ import { type Page, pages } from './pages/index.ts'
 import { type Route, router } from './router.ts'
 
 /**
- * Iterates over all parameters of a page, calling the iterator function 
+ * Iterates over all parameters of a page, calling the iterator function
  * for each parameter store and its corresponding value from the route
  */
 function eachParam<SomeRoute extends Route>(


core/loader/index.ts
```diff
@@ -5,10 +5,26 @@ import { jsonFeed } from './json-feed.ts'
...
+  /**
+   * Given the website html response, returns the default feed url,
+   * e.g. https://example.com/rss for rss feed.
+   */
   getSuggestedLinksFromText(response: TextResponse): string[]
+  /**
+   * Returns feed title, if any
+   */
   isMineText(response: TextResponse): false | string
+  /** TODO What does this do or should do? Not too clear from tests. */
```
ai, line 27: For instance, YouTube loader will return `true` for `youtube.com` links or Telegram loader for `t.me` links.

It detects that URL is 100% for this loader. It is not used right now because there is no way to detect RSS/Atom link just by URL.

core/loader/index.ts
```diff
@@ -29,6 +45,10 @@ export interface FeedLoader {
   url: string
 }
 
+/**
+ * Decides which loader to use for the given feed response.
+ * TODO Looks like the loader is always atom
```
ai, line 50: Hm, why?

PR #303, opened 2025-06-12
Atom pagination support

### LINKED ISSUE #56: Add Atom pagination support
Atom feed could be not only on single page, but with pagination.

```xml
<link rel="next" href="http://www.syfyportal.com/atomFeed.php?page=4"/>
```

https://www.rfc-editor.org/rfc/rfc5005#section-3

We page `PostPage` to handle pagination well, so we need just to update loader and add tests.

ashlkv:
Fixes #56

Adds atom pagination support in Atom loader. Taps into `getPosts` existing logic, which already returns a tuple of posts + post loader. And if there is a link to next posts in the feed, returns the loader for the next posts recursively.

# Motivation

While the pagination is not too widely used in blog feeds (roughly 1.6%), it's a nice-to-have for Mastodon and other social media.

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -7,6 +7,7 @@ _See the [full architecture guide](../README.md) first._
 - [Client Environments](#client-environments)
 - [URL Routing](#url-routing)
 - [Test Strategy](#test-strategy)
+  - [Mocking Requests](#mocking-requests)
 
 ## Project Structure
 
@@ -87,7 +88,7 @@ In VS Code you can use [extension](https://marketplace.visualstudio.com/items?it
 
 Open `core/coverage/lcov-report/index.html` to see coverage issues.
 
-## Mocking requests
+### Mocking Requests
 
 To enable network request mocking in tests, you have to set up and tear down request mock before and after each test:
 
--- a/core/README.md
+++ b/core/README.md
@@ -88,7 +88,7 @@ In VS Code you can use [extension](https://marketplace.visualstudio.com/items?it
 
 Open `core/coverage/lcov-report/index.html` to see coverage issues.
 
-### Mocking Requests
+## Mocking Requests
 
 To enable network request mocking in tests, you have to set up and tear down request mock before and after each test:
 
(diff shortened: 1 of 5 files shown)


core/test/loader/atom.test.ts
```diff
@@ -457,3 +467,118 @@ test('parses media', () => {
...
+    )
+  )
+  equal($store.get().hasNext, false)
+})
+
+test('loads first then second page', async () => {
+  let task = createDownloadTask()
+
+  let callCount = 0
+  let textSpy = spyOn(task, 'text', () => {
```
ai, line 503: We have our own HTTP request mocking system which can be used here.

You can find an example here: https://github.com/hplush/slowreader/blob/main/core/test/pages/add.test.ts

You need to replace `task.text()` to `expectRequest()` call before the check.

core/loader/atom.ts
```diff
@@ -36,6 +37,69 @@ function parsePosts(text: TextResponse): OriginPost[] {
...
+  feedUrl: string,
+  feedResponse: TextResponse | undefined
+): PostsCursor {
+  if (!feedResponse) {
+    return [
+      undefined,
+      async () => {
+        let response = await task.text(feedUrl)
+        let [posts, loader] = getPostsCursor(task, feedUrl, response)
+        return [posts, loader] as [OriginPost[], PostsListLoader | undefined]
```
ai, line 76: Why do we need this type cast? Maybe we can just add extra `if` like `if (post && loader)`?

core/test/loader/atom.test.ts
ashlkv: Done and added a section on request mocking to readme.

core/README.md
```diff
@@ -86,3 +86,22 @@ n bnt core/test/html.test.ts -t 'sanitizes HTML'
 In VS Code you can use [extension](https://marketplace.visualstudio.com/items?itemName=connor4312.nodejs-testing) to run specific test from UI.
 
 Open `core/coverage/lcov-report/index.html` to see coverage issues.
+
+## Mocking requests
```
ai, line 90: Nice! I really like your passion of long-term solution (like updating docs to help other to find mocks).

Just let’s move it to `core/README.md` because it works only there

ashlkv: It's already in `core/README.md` or?

core/loader/atom.ts
ashlkv: Yep let's avoid type cast.

core/README.md
ai: You right, sorry :D. I am on the conference, seems like it is easy to misread :D

PR #310, opened 2025-07-24
feat:CopyButton-to-Output-Control

### LINKED ISSUE #306: Add Copy button to `<Output>` control
We use `<Output>` control on `Signup` page to allow user to copy random user ID / secret. We will have better UX if we will have a copy button directly.

1. Start visual tests and open Output story.
2. Edit `web/ui/output.svelte`
3. Add `size="icon"` button with copy icon.
4. After click change icon to OK icon for a few seconds (use `aria-live` to announce `copied` for screen readers).
5. Copy text by using [async Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API).
6. Add OK state to Output story.
7. Open Signup page story and check that everything works.

suyash5053:
Fixes #306 

Added a copy button to the `<Output>` control with the proper size and aria-live functionality. Let me know if any changes

# Screenshot or Video
https://www.loom.com/share/456fc0c895744877860105c251cc6929?sid=7672b27d-6bd9-4d0e-bf42-c50ca89ea8d7

### DIFF
--- a/web/ui/output.svelte
+++ b/web/ui/output.svelte
@@ -1,6 +1,8 @@
 <script lang="ts">
+  import { mdiCheck, mdiContentCopy } from '@mdi/js'
   import type { HTMLInputAttributes } from 'svelte/elements'
 
+  import Button from './button.svelte'
   import Label from './label.svelte'
 
   let {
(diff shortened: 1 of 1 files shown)


web/ui/output.svelte
```diff
@@ -14,20 +16,59 @@
...
+    />
+    <Button
+      class="output_button"
+      aria-label={isCopied ? 'Copied' : 'Copy to clipboard'}
+      icon={currentIcon}
+      onclick={copyToClipboard}
+      size="icon"
+      variant="main"
+    >
+      <span class="">Copy to clipboard</span>
```
ai, line 64: Why we need extra `<span>`?

ai:
It will be nice to have storybook test for that. Just add another `<Output>` to `web/stories/ui/output.stories.svelte` and write a JS there to click this `button` every 3 seconds.

web/ui/output.svelte
suyash5053: Actually I gave it some styling, and then I thought since Its not going to be visible its worthless, removed the class but left the span. `console.log("Junior Engineer Issues")`

ai: It happens. I have ESLint `no-console` because without it always push changes with debug’s `console.log` :D

ai:
> To be clear this is what we want?

Much closer. But it is important to not have a gap between right border of `<Output>` and `<Button>`

470271847-b89c93b4-59c9-4870-a894-89b105454db5

ai:
Also, we may improve background colors a little. 

1. Move background from `.output_field` to `.output`
2. Add `--current-background: the-same-bg` to `.output`

In this case `<Button>` colors will be calculated from the output’s background. We use relative colors a lot (in button too) and here is one of example.

web/ui/output.svelte
```diff
@@ -38,6 +38,24 @@
       throw e
     }
   }
+
+  export async function triggerCopy(): Promise<void> {
```
ai, line 42: It is a bad practice creating a special code for test because in storybook you test it and not the real behavior.

web/ui/output.svelte
```diff
@@ -14,43 +16,121 @@
...
+      position: relative;
       margin: 0.625rem 0;
+
+      --current-background: oklch(
+        from var(--current-background) calc(l + var(--field-l))
+          calc(c + var(--field-c)) h
+      );
+
+      .output_button button[aria-label='Copied'] svg {
+        --success-color: oklch(0.7227 0.192 149.58);
```
ai, line 105: We try to keep all colors in `web/main/colors.css`

Also, try to trim digits `oklch(0.7227 0.192 149.58)` → `oklch(0.70 0.2 150)` (or similar depending on how it looks like). Too many digits in the number just increase CSS bundle without any visual changes.

web/ui/output.svelte
```diff
@@ -14,43 +16,121 @@
...
     .output {
+      position: relative;
       margin: 0.625rem 0;
+
+      --current-background: oklch(
+        from var(--current-background) calc(l + var(--field-l))
+          calc(c + var(--field-c)) h
+      );
+
+      .output_button button[aria-label='Copied'] svg {
```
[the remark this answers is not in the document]
ai: Also, you don’t need to write `.output_button` inside `.output`. The selector already has `.output` prefix. Move it to `:global`.

If you have nested rules, they will have different CSS selectivity which will force you to use `!important` in some future.

web/ui/output.svelte
suyash5053: Got it

PR #321, opened 2025-12-06
Improve SPA UI integration to the system

ai:
Makes SPA UI from border to border (and add safe paddings for OS notifications icons on the top and bar on the bottom)


ai:
We still have black bottom bar on Android

PR #326, opened 2026-01-03
Support If-Modified-Since header for feed caching

### LINKED ISSUE #324: Use `If-Modified-Since` to speed up feed update
* Add `refreshedAt` to `FeedValue`
* Write it to feed in `core/refresh.ts`
* Send time as [`If-Modified-Since`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-Modified-Since) with HTTP request updating the feed
* Do not parse and check content if server responded with `304` status.

immitsu:
Resolves #324

### DIFF
--- a/core/loader/common.ts
+++ b/core/loader/common.ts
@@ -190,7 +190,7 @@ export function findMediaInText(
 }
 
 /**
- * Fetches feed data if modified, handling conditional requests.
+ * Prevent heavy loading/parsing by `If-Modified-Since` header.
  */
 export async function fetchIfModified(
   task: DownloadTask,


immitsu:
If everything is generally ok, we can add `core/test/feed.test.ts` here to restore test coverage

proxy/index.ts
```diff
@@ -126,6 +126,20 @@ export function createProxy(
         /* node:coverage enable */
       }
 
+      if (
```
ai: Nice, I like it.

I just a little afraid of broken `last-modified` header will block us from using resource. Maybe in case of parsing error print a warning and send targetr reponse?

immitsu: Done

core/feed.ts
```diff
@@ -99,6 +100,24 @@ export function getFeedLatestPosts(
   feed: FeedValue,
   task = createDownloadTask()
 ): PostsList {
+  if (feed.refreshedAt) {
+    let originalTask = task
+    task = {
```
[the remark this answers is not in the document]
immitsu: Please take a look

core/test/loader/rss.test.ts
```diff
@@ -449,3 +450,79 @@ test('returns post source', async () => {
...
+        originId: 'https://example.com/1',
+        publishedAt: undefined,
+        title: 'Test Post',
+        url: 'https://example.com/1'
+      }
+    ]
+  })
+})
+
+test('handles 304 not modified response', async () => {
```
ai: I am trying to avoid “unit” test ideology in the project. My dream is to test behavior, not implementation.

What do you think of test scenarios like:

1. Mock response and return `Last-Modified` header
2. Run `refresh()`
3. Mock second response and expect to have `If-Modified-Since` and response with 304
4. Run `refresh()` again

With this scenario we will check that the whole system works, and we will not need to change test on any implementation changes (like on changing Loader API and `Last-Modified` storage).

immitsu: Reworked

core/test/loader/rss.test.ts
```diff
@@ -449,3 +450,112 @@ test('returns post source', async () => {
...
+        originId: 'https://example.com/1',
+        publishedAt: undefined,
+        title: 'Test Post',
+        url: 'https://example.com/1'
+      }
+    ]
+  })
+})
+
+test('handles conditional request when server returns 304', async () => {
```
ai: I like new tests.

The only note: what do you think if we will have the same 1 test here as we have for Atom/JSONFeed (or maybe I am missing something)?

immitsu:
> Nice, I like new feature a lot =^_^= Thanks for your awesome work.

Thank you for this project (and many others), the detailed review, and the valuable feedback

core/loader/atom.ts
```diff
@@ -122,9 +122,15 @@ function parseFeed(
 
 async function loadFeed(
   task: DownloadTask,
-  url: string
+  url: string,
+  refreshedAt?: number
 ): Promise<PostsListResult> {
-  return parseFeed(task, await task.text(url))
+  let headers = refreshedAt
```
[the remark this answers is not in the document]
immitsu: Added

core/test/loader/rss.test.ts
ai: [Recent Danny Preussler talk](https://www.droidcon.com/2024/10/17/ten-things-you-heard-about-testing-that-might-be-wrong/) convinced me that unit tests approach is not very helpful. The price of test maintenance is more that it gives in return because we need constantly update tests.

Separated tests for like this is that “unit test” approach. For instance, if we will change loader API, we will need to update test.

The new idea is that tests should test “user side feature” and not implementation (”If it walks like a duck and quacks like a duck, it's a duck”).

So I suggest removing one test and instead have only one test (technically 3, one for Atom, one for RSS, one for JSON Feed) that check both thing.

It will be nice to have in `refresh.test.ts` another test, which will be like e2e test where we:
1. Create feed
2. Mock request with 200 and time header and test that it was without header
3. Refresh posts
4. Mock request with 304 and check that time from step 2 was sent
5. Refresh post again

This test will test that app saves feed’s last updated time somewhere and use it again (but do not care about implementation).

Of course, we can do this test in separated PR (or I can generate it with LLM if you don't like tests :D). But I want to share with you this idea of more e2e-ish approach with testing.

core/loader/common.ts
```diff
@@ -183,3 +188,20 @@ export function findMediaInText(
     } satisfies PostMedia
   })
 }
+
+/**
+ * Fetches feed data if modified, handling conditional requests.
```
ai:
```suggestion
 * Prevent heavy loading/parsing by `If-Modified-Since` header.
```

PR #331, opened 2026-01-20
Fix duplicate feed links extracted from HTTP pages

ilyatitovich:
- Added `dedupeLinks` function to ensure that feed links differing only by protocol are not duplicated;
- Added a test.

### Motivation

When a user provides an HTTP URL, the fetched HTML may reference the same feed using both absolute HTTPS links and relative links resolved from the HTTP page. This resulted in duplicate feed candidates being returned.

### Note

For manual testing, you can use [this page](https://ilyatitovich.github.io/test-feeds/). It contains three potential feeds that would appear as duplicates if you use HTTP without this fix.

### DIFF
--- a/core/pages/add.ts
+++ b/core/pages/add.ts
@@ -147,19 +147,34 @@ export const addPage = createPage('add', () => {
   })
 
   /**
+   * Removes duplicated feed links differing only by protocol.
+   */
+  function dedupeLinks(links: string[]): string[] {
+    let seen = new Set<string>()
+    return links.filter(link => {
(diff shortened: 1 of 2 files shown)


ilyatitovich:
Maybe I misunderstood, but I remember how I found this small bug. We discussed it on Discord: if you open a site via HTTP, the feed link extracted from the anchor appears in the list as http, creating a duplicate of the same link that already exists as https. So I assumed that only this behavior needed to be fixed.

Now I need to think about this subscription issue 😀

PR #333, opened 2026-01-28
Add subprotocol check to the client

### LINKED ISSUE #245: Add subprotocol check to the client
User may use old client. We need to handle this case and instead of server error (because of old protocol) show the proper message to update client.

To do it we use “Subprotocol”, client-server API version. We use it in Logux. Logux clients send their version in all actions. Logux Server has a range of supported versions.

We need to add the same logic to HTTP API:

1. Add `X-Subprotocol` header to [`fetchJSON()` helper](https://github.com/hplush/slowreader/blob/main/api/http/utils.ts)
2. See [Logux Server](https://github.com/logux/server/blob/main/server-client/index.js#L119-L126) code to find our how to check supported range.
3. Add this check to [`jsonApi()`](https://github.com/hplush/slowreader/blob/main/server/lib/http.ts) and respond a proper error message on it.
4. Add a `Update client` route to the `core/` router. You may need boolean store and add logic to `compute` in the router.
5. Add a `Update client` to the web client with simple design. See [How create a page](https://github.com/hplush/slowreader/blob/main/docs/new_page.md).
6. Add code to the core to set that boolean store to `true` in case of Logux error or `fetchJSON` error, which ask to update the client.

ilyatitovich:
Fixes #245  

* Added `isClientUpdateRequired` atom store to trigger **Update Client page** when the client uses an unsupported subprotocol (below server `minSubprotocol`);
* Added handling of unsupported subprotocols for both **HTTP requests** and **Logux WebSocket errors**;
* Added router logic to redirect users to `/update-client` when a client update is required;
* Added a simple **Update Client page** with temp test data;
* Added tests.

### Motivation

This change introduces a unified mechanism for detecting unsupported client versions across HTTP and Logux, and redirects users to an update page.

### DIFF
--- a/core/pages/update-client.ts
+++ b/core/pages/update-client.ts
@@ -1,10 +1,11 @@
-import { isOutdatedClient } from '../client.ts'
 import { createPage } from './common.ts'
 
 export const updateClientPage = createPage('updateClient', () => {
+  /* node:coverage disable */
   function handleUpdateClient(): void {
-    isOutdatedClient.set(false)
+    location.reload()
   }
+  /* node:coverage enable */
 
   return {
     exit() {},
--- a/web/main/main.svelte
+++ b/web/main/main.svelte
@@ -2,8 +2,10 @@
   import {
     busy,
     currentPage,
+    isOutdatedClient,
     layoutType,
     notFound,
+    pages,
     popupsStatus,
     signOut,
     subscribeUntil,
@@ -60,6 +62,8 @@
   {#if !globalLoader}
     <BusyPage />
   {/if}
+{:else if $isOutdatedClient}
+  <UpdateClientPage page={pages.updateClient()} />
 {:else if $currentPage.route === 'notFound'}
   <NotFoundPage />
 {:else if $currentPage.route === 'fast'}
@@ -86,8 +90,6 @@
   <ExportPage page={$currentPage} />
 {:else if $currentPage.route === 'import'}
   <ImportPage page={$currentPage} />
-{:else if $currentPage.route === 'updateClient'}
-  <UpdateClientPage page={$currentPage} />
 {:else}
   <ThinPage title={$currentPage.route}>
     {$currentPage.route}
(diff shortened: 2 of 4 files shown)


server/lib/http.ts
```diff
@@ -67,6 +67,24 @@ export function jsonApi<Response, Request extends object>(
...
+      let clientSubprotocol = Number(req.headers['x-subprotocol'])
+
+      if (
+        isNaN(clientSubprotocol) ||
+        clientSubprotocol < server.options.minSubprotocol
+      ) {
+        res.writeHead(400, {
+          'Access-Control-Expose-Headers': 'X-Client-Action',
+          'Content-Type': 'text/plain',
+          'X-Client-Action': 'update-client'
```
ai: Let’s use response body.

Custom header is a valid option, but we're already using response body as API so it will be more consistent.

Here is an example:
1. Defining string token in API https://github.com/hplush/slowreader/blob/main/api/http/signup.ts#L47
2. Using it in the server https://github.com/hplush/slowreader/blob/1e1c62e711c302057366d28181d38a58f3d5359c/server/modules/auth.ts#L71
3. On the client we can compare answer with the token from the same file

web/main/main.svelte
```diff
@@ -85,6 +86,8 @@
   <ExportPage page={$currentPage} />
 {:else if $currentPage.route === 'import'}
   <ImportPage page={$currentPage} />
+{:else if $currentPage.route === 'updateClient'}
```
ai: No, the new idea is to remove `isOutdatedClient` from `core/router.ts` and instead have `{:else if $isOutdatedClient}` before all pages in `main.svelte`

core/pages/update-client.ts
```diff
@@ -0,0 +1,16 @@
+import { isOutdatedClient } from '../client.ts'
+import { createPage } from './common.ts'
+
+export const updateClientPage = createPage('updateClient', () => {
+  function handleUpdateClient(): void {
+    isOutdatedClient.set(false)
```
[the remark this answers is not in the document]
ilyatitovich: As a temporary solution, this looks fine. I remember that we avoid putting platform-specific code into `core/`.

Do we want to keep this page in `core/pages/` if it only uses a single function? Maybe it makes sense to move this function to `core/client.ts` and just import it in the page.

PR #338, opened 2026-05-12
Lazy load rarely used pages

### LINKED ISSUE #38: Split JS bundle
Some pages are rarely used. We should speed-up app starting time by avoiding parsing JS of that pages.

We will use per-page splitting.

- [ ] Add `page loading` page
- [ ] Split pages to groups according to how often user will use them together
- [ ] Split JS bundle

mugnivenko:
Fixes #38 

I added lazy loading of the about, export, and import pages.

# Motivation

To make the size of the loading JS smaller.

### DIFF
--- a/web/.size-limit.json
+++ b/web/.size-limit.json
@@ -16,13 +16,13 @@
       "!dist/assets/noto-sans-mono-*.woff2",
       "!dist/assets/playpen-sans-*.woff2",
       "!dist/assets/*-{italic,math,latin-ext}-*.woff2",
-      "!dist/assets/{export,import,about}-*.js"
+      "!dist/assets/{settings}-*.js"
     ],
     "limit": "190 KB"
   },
   {
     "name": "Core scripts to execute",
-    "path": "dist/assets/index-*.js",
+    "path": "dist/assets/{index,core}-*.js",
     "brotli": false,
     "limit": "310 KB"
   }
--- a/web/.size-limit.json
+++ b/web/.size-limit.json
@@ -15,7 +15,8 @@
       "!dist/assets/noto-*-{cyrillic,greek,devanagari,vietnamese}-*.woff2",
       "!dist/assets/noto-sans-mono-*.woff2",
       "!dist/assets/playpen-sans-*.woff2",
-      "!dist/assets/*-{italic,math,latin-ext}-*.woff2"
+      "!dist/assets/*-{italic,math,latin-ext}-*.woff2",
+      "!dist/assets/{export,import,about}-*.js"
     ],
     "limit": "190 KB"
   },
--- a/web/main/main.svelte
+++ b/web/main/main.svelte
@@ -10,15 +10,12 @@
     userId
   } from '@slowreader/core'
 
-  import AboutPage from '../pages/about.svelte'
   import AddPage from '../pages/add.svelte'
   import BusyPage from '../pages/busy.svelte'
   import CloudPage from '../pages/cloud.svelte'
   import DownloadPage from '../pages/download.svelte'
-  import ExportPage from '../pages/export.svelte'
   import FeedByCategoriesPage from '../pages/feeds-by-categories.svelte'
   import FeedsPage from '../pages/feeds/index.svelte'
-  import ImportPage from '../pages/import.svelte'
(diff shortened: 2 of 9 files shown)


web/main/main.svelte
```diff
@@ -76,7 +73,9 @@
 {:else if $currentPage.route === 'feedsByCategories'}
   <FeedByCategoriesPage page={$currentPage} />
 {:else if $currentPage.route === 'about'}
-  <AboutPage page={$currentPage} />
+  {#await import('../pages/about.svelte') then {default: AboutPage}}
```
ai: I think we forgot about some loader during the page loading.

web/main/main.svelte
ai: As I mention in the issue it is better strategy to join pages in the bundles according to user behavior.

I think we can bundle all settings (expect `import` and `export`) into the single bundle.

web/.size-limit.json
```diff
@@ -15,7 +15,8 @@
       "!dist/assets/noto-*-{cyrillic,greek,devanagari,vietnamese}-*.woff2",
       "!dist/assets/noto-sans-mono-*.woff2",
       "!dist/assets/playpen-sans-*.woff2",
-      "!dist/assets/*-{italic,math,latin-ext}-*.woff2"
+      "!dist/assets/*-{italic,math,latin-ext}-*.woff2",
+      "!dist/assets/{export,import,about}-*.js"
```
ai: Maybe by moving to custom bundles we can add some custom prefix?

web/.size-limit.json
ai, line 27: Let’s update core scripts.

But as I see `button-*` and `core-*` was moved out. Do we need this bundle to show home page? If yes, we need to update `Core scripts to execute` metrics (and it is better to have some automation to avoid manually picking bundles).

mugnivenko: I've updated the script according to the chunks

image

web/pages/settings/index.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang="ts">
```
ai, line 1: Pages folder is not beautiful page right now. JS bundle splitting is a simple feature which we should not think about, it should not force us to change architecture and project stricture.

web/main/main.svelte
```diff
@@ -75,16 +72,16 @@
   <AddPage page={$currentPage} />
 {:else if $currentPage.route === 'feedsByCategories'}
   <FeedByCategoriesPage page={$currentPage} />
-{:else if $currentPage.route === 'about'}
-  <AboutPage page={$currentPage} />
-{:else if $currentPage.route === 'cloud'}
-  <CloudPage page={$currentPage} />
+{:else if $currentPage.route === 'about' || $currentPage.route === 'interface' || $currentPage.route === 'cloud'}
```
ai, line 75: 3 pages moved to JS chunk is still almost nothing. We need solid and finished solution for all pages.

You need to think about good strategy first, do not rush.

mugnivenko:
I have thought about how to keep the feature out of developers’ concerns and make it work automatically, but I couldn’t come up with a solution. Either we need to somehow combine pages into the desired chunks in the code or make all pages async and combine them into chunks via Vite’s `manualChunks` option. Both solutions require developers to think about code splitting every time they add a new page. I’m out of options in my understanding of the problem. Can you please give possible directions to explore for new solutions?

ai:
The whole task is about finding the right way to implement. Let’s move task back to waiting list (maybe new API will be introduced)

mugnivenko:
To introduce the new API I need clearer requirements about what the final result should look like. The requirements now quite contradict each other.
>  We should not touch the project structure and think about splitting.

As we do not use a framework that gets the work done for us, we should at least think about code splitting in the `main.svelte` file.
The desire to combine some pages into one loadable chunk makes us somehow introduce this requirement in the code.
As a possible solution, I may suggest writing a Vite plugin. The plugin will automatically add lazy loading for all pages and will use a custom config that defines which pages to combine. The solution might be overkill for this case, but it at least covers all requirements.

ai:
I told about Vite API.

mugnivenko:
Could you please point me to where you mentioned the Vite API? I couldn’t find it. Also, do you consider this solution suitable?

PR #340, opened 2026-08-27
Add copy button to <Output> for easier value reuse

### LINKED ISSUE #306: Add Copy button to `<Output>` control

batyshkaLenin:
Fixes #306

Adds an inline copy button to `<Output>` using the async Clipboard API. After a successful copy, it shows a green check for 3 seconds and announces the state with `aria-live`.

Adds translated labels and a Storybook `Copied` state driven by a real click.

Incorporates the review feedback from #310. Thanks @suyash5053 for the earlier work.

# Motivation

Make generated credentials and other read-only values easier to copy.

# Screenshot or Video

|                                                                           Light                                                                            |                                                                           Dark                                                                            |
|:----------------------------------------------------------------------------------------------------------------------------------------------------------:|:---------------------------------------------------------------------------------------------------------------------------------------------------------:|
| Copied Output in light theme  | Copied Output in dark theme  |
| Copied Output in light theme  |            image             |
| Copied Output in light theme | Copied Output in dark theme |
| Copied Output in light theme  | Copied Output in dark theme  |

[Video in light theme](https://github.com/user-attachments/assets/bf7ee7c2-d180-4077-822c-70bb8d736b98)


ai:
Thanks for the screenshot. I think we need a little more work for how it looks and feel.

I think we need to change button design a little. The border radius on the left looks weird.

And the overall physics of design is uncomfortable. The button looks more deep (into z-axis) than the field. It is strange to press something so deep inside. Our common sense tells us that button is often upfront to have some space for the press.

web/ui/output.svelte
```diff
@@ -14,39 +20,94 @@
...
+      {...props}
+    />
+    <div class="output_button" class:is-copied={copied}>
+      <Button
+        icon={copied ? mdiCheck : mdiContentCopy}
+        onclick={copy}
+        size="icon"
+        variant="secondary"
+      >
+        {copied ? $t.copied : $t.copyToClipboard}
```
ai, line 65: It would be nice to announce `Copied` by `aria-live`

ai:
Wait a second, You showed presses state in the PR description?

In this case the z-position is OK, but we need to fix an issue with the border radius.

batyshkaLenin:
@ai Yes, that’s the pressed state. I created a page in Storybook showing the pressed state and took the screenshots from there.

web/ui/output.svelte
batyshkaLenin: Copied is already announced by `<Announce text={copied ? $t.copied : ''} />` below the control. Announce renders role="status" with aria-live="polite".

ai: Nice! Sorry, I missed it

batyshkaLenin:
Maybe something like this? I haven’t committed it yet - I’m still thinking about the best way to do it.

|                                                                           Light                                                                            |                                                                           Dark                                                                            |
|:----------------------------------------------------------------------------------------------------------------------------------------------------------:|:---------------------------------------------------------------------------------------------------------------------------------------------------------:|
| image  | image  |
| image | image |
| image | image |

ai:
I like it :+1:

web/ui/output.svelte
```diff
@@ -14,39 +20,99 @@
...
+    .output_button {
+      --current-background: var(--output-background);
+
+      position: absolute;
+      inset-block-start: 0;
+      inset-inline-end: 0;
+    }
+
+    .output_button > button {
+      border-start-start-radius: 0;
```
ai: Sorry is too hacky and step over CSS isolation.

We need to add some attribute to `Button`. For instance, `joined={"start" | "end" | "middle" | undefined}`

ai:
Thanks!

batyshkaLenin:
Thanks for the recommendations regarding isolation. I write client-side applications very rarely, so my frontend skills have gotten a bit rusty. I figured that if I’m going to refresh them, I might as well do it properly.

I have another question: do you know of any good articles about working with Storybook, apart from the official recommendations? While working on this task, I started wondering whether I’m organizing the components in Storybook correctly.

ai:
> I have another question: do you know of any good articles about working with Storybook

Sorry, no.

I think even my way is not a canonical way (my way is copy from my own “Storybook” which I used before Storybook became popular).

My way:
1. Try to test pages and app states, not just components like buttons
2. For components put as many components on the screen to see issues in one screen (also to save money of paid Chromatic)

